PR #7, opened 2017-03-25
Port to py27 again

### LINKED ISSUE #5: Is there interest/value in making cattrs work on python 2.7?
* cattrs version: 0.2.0
* Python version: 2.7.13
* Operating System: MacOS Sierra 10.12.2

### Description

I wanted to try to use cattrs on python 2.7 for a personal project
I know that this is supposed to be a python3-specific project, but ...

### What I Did

... I made a few changes, and was able to create a [branch](https://github.com/adetokunbo/cattrs/tree/port-to-py27/cattr) that successful runs the current tests on python 2.7.  I.e

```
tox -e py27
```

works.

There's still work to do:

- the changes have broken python3.5, though that looks to be due to differences between what's available in the typing [backport on pypi](https://pypi.python.org/pypi/typing) and what's available in python3.5, rather than changes needed to make cattrs work on py2.7

I can work on fixing the issues and submit a PR, but I wanted to get early feedback before spending time on it.  E.g, it'd be worth knowing if cattrs should support python2.7 or not.

adetokunbo:
Fixes #5.

With this series of commits, the cattrs should work on python 2.7.
Each commit message explains the reason for its change.  
In summary:
- the unittests pass largely unchanged with minor tweaks to account for python2/3 differences
- typing for python2.7 needed to be vendor and patched
- the only significant change modification to converter was to a change to how the unstruct_strat attribute is used


adetokunbo:
:) The large negative coverage drop comes from having to vendor the python2 version of typing.py

Tinche:
Hi!

Thanks for the great work. Some questions:

I see the ``typing`` module on PyPI is dated 2017/01/01, so I'm guessing that's the reason we're vendoring it for Python 2 too? If we could get someone to update the PyPI package, could be stop vendoring it for Python 2?

In any case, I don't really see the point of including vendored stuff in our test coverage reports. If you take a look at setup.cfg, our vendored ``typing`` is already blacklisted, so please fix the paths and add the Python 2 file too.

I see you've changed UnstructureStrategy to not be an explicit str enum. I think I'm okay with that.

You've removed the getter and setter for ``Converter.unstruct_strat``. The tests seem to pass fine using this approach, and it's a little faster than the approach in the PR, so could you please restore that?

PEP 484 suggests placing the comment type signature for functions before the docstring (https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code), so please do it this way.

That's all I can think of! Very close to merging this in :)

adetokunbo:
> I see the typing module on PyPI is dated 2017/01/01, so I'm guessing that's the reason we're vendoring it for Python 2 too? If we could get someone to update the PyPI package, could we stop vendoring it for Python 2?

No, unfortunately there is an additional issue that requires a patch - I describe it in python/typing#405

> In any case, I don't really see the point of including vendored stuff in our test coverage reports. If you take a look at setup.cfg, our vendored typing is already blacklisted, so please fix the paths and add the Python 2 file too.

Fixed

> You've removed the getter and setter for Converter.unstruct_strat. The tests seem to pass fine using this approach, and it's a little faster than the approach in the PR, so could you please restore that?

Fixed, via a workaround to make it work in python2

The reason I made this change was that in python2, invoking the setter in `__init__.py` did not work, and the Converter instance ended up not having an unstructure_attrs method:

```python
    def __init__(self, ..., unstruct_strat=UnstructureStrategy.AS_DICT):
        ...
        self.unstruct_strat = unstruct_strat  # this does not trigger the setter method in python2.7

    @property
    def unstruct_strat(self):
         ...
    @unstruct_strat.setter		
    def unstruct_strat(self, val):
        ...	
```

> PEP 484 suggests placing the comment type signature for functions before the docstring (https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code), so please do it this way.

Fixed

Tinche:
Alright, look great to me. I'll merge it in in a day or two.

> The reason I made this change was that in python2, invoking the setter in __init__.py did not work, and the Converter instance ended up not having an unstructure_attrs method:

That's kind of weird. It works for me (Python 2.7.12 on Ubuntu).

Tinche:
In this goes! Thanks for your hard work.

adetokunbo:
Great! It was a pleasure.

Issue #9, opened 2017-08-31
Switch to registering functions vs singledispatch


toumorokoshi:
* cattrs version: 0.4.0

### Description

I've run into a couple issues with using cattrs, and I think switching from singledispatch to a combination of can_handle_type functions and a lrucache will fix them.

Here's the problems I've found:

1. importing typing and cattrs in the same python interpreter leads to an infinite recursion error for Python2.

I understand the motivation for vendor typing, but it makes it impossible to use a project with the standard typing module and cattrs. This example in Python2 leads to an recursive error, recursing in the vendored typing forever.

```
import attr
import cattr
from typing import List


@attr.s
class Point(object):
    x = cattr.typed(int)
    y = cattr.typed(float)


point = cattr.structure({"x": 1, "y": 1.0}, Point)

def foo():
    return [1, 2, 3]
```

2. The vendored typing is old

The vendored typing module doesn't contain some fixes, such as: 

https://bugs.python.org/issue28339

Leading to an error message like:

https://travis-ci.org/toumorokoshi/transmute-core/jobs/269864064

```
  File "/home/travis/build/toumorokoshi/transmute-core/lib/python3.5/site-packages/_pytest/config.py", line 379, in _importconftest
    mod = conftestpath.pyimport()
  File "/home/travis/build/toumorokoshi/transmute-core/lib/python3.5/site-packages/py/_path/local.py", line 662, in pyimport
    __import__(modname)
  File "/home/travis/build/toumorokoshi/transmute-core/transmute_core/__init__.py", line 12, in <module>
    from .swagger import generate_swagger_html, get_swagger_static_root, SwaggerSpec
  File "/home/travis/build/toumorokoshi/transmute-core/transmute_core/swagger/__init__.py", line 1, in <module>
    import jinja2
  File "/home/travis/build/toumorokoshi/transmute-core/lib/python3.5/site-packages/jinja2/__init__.py", line 33, in <module>
    from jinja2.environment import Environment, Template
  File "/home/travis/build/toumorokoshi/transmute-core/lib/python3.5/site-packages/jinja2/environment.py", line 15, in <module>
    from jinja2 import nodes
  File "/home/travis/build/toumorokoshi/transmute-core/lib/python3.5/site-packages/jinja2/nodes.py", line 19, in <module>
    from jinja2.utils import Markup
  File "/home/travis/build/toumorokoshi/transmute-core/lib/python3.5/site-packages/jinja2/utils.py", line 486, in <module>
[...]

PR #11, opened 2017-09-16
adding a function based dispatch.

toumorokoshi:
This address #9.

A proof of concept PR, so I could do more polish on it. But it illustrates:

1. using a FunctionDispatch class vs SingleDispatch.
2. feature parity with SingleDispatch (including caching for quick lookup)

Tried it out locally, works with py27 and py36, sans the vendor typing library.

More improvements / changes:

- change the register function signature to take two functions, rather than type + function.
- use FunctionDispatch to also handle attrs

### DIFF
--- a/cattr/multistrategy_dispatch.py
+++ b/cattr/multistrategy_dispatch.py
@@ -11,10 +11,10 @@ class MultiStrategyDispatch(object):
     """
 
     def __init__(self, fallback_func):
+        self._dispatch_not_found = object()
         self._function_dispatch = FunctionDispatch()
         self._function_dispatch.register(lambda cls: True, fallback_func)
-        self._single_dispatch = singledispatch(lambda: None)
-        self._clear_registry(self._single_dispatch)
+        self._single_dispatch = singledispatch(self._dispatch_not_found)
         self._cache = {}
 
     @staticmethod
@@ -36,9 +36,9 @@ class MultiStrategyDispatch(object):
             found = False
             try:
                 dispatch = self._single_dispatch.dispatch(cl)
-                if dispatch is not None:
+                if dispatch is not self._dispatch_not_found:
                     found = True
-            except Exception:
+            except Exception as e:
                 pass
             if not found:
                 dispatch = self._function_dispatch.dispatch(cl)
@@ -48,7 +48,13 @@ class MultiStrategyDispatch(object):
     def register_cls(self, cls, handler):
         """ register a class, which utilizes singledispatch """
         self._single_dispatch.register(cls, handler)
+        if cls in self._cache:
+            del self._cache[cls]
 
-    def register_func(self, func, handler):
-        """ register a function to determine if the handle should be used for the type """
-        self._function_dispatch.register(func, handler)
+    def register_func_list(self, func_and_handler):
+        """ register a function to determine if the handle
+            should be used for the type
+        """
+        for func, handler in func_and_handler:
+            self._function_dispatch.register(func, handler)
+        self._cache.clear()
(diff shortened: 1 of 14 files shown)


Tinche:
Thanks! I'll take a look soon.

Tinche:
The problem with this approach is that the following snippet stops working:
```
import attr
import cattr


@attr.s
class A:
    a = attr.ib()


@attr.s
class B(A):
    b = attr.ib()


cattr.register_structure_hook(B, (lambda d, t: B(**d)))
cattr.register_structure_hook(A, (lambda d, t: A(**d)))

print(cattr.structure({'a': 1, 'b': 2}, B))
```

I'm OK with having a list of matchers that are consulted in turn when looking up a handler function (and we will probably require this to handle Unions with parameters), but the last entry in this list should be something that understands the class hierarchy and the MRO. Then the API that prepends to the list would be a low level kind of API, and just registering a type in the last entry would be a high level kind of API.

toumorokoshi:
Yeah, that approach sounds reasonable: best of both worlds. 

Probably will work with singledispatch at the end, with a caveat on it's use on Typing objects. I'll give it a try.

toumorokoshi:
I'll add some more coverage for the hooks, but my general approach is done. I'll probably start commenting now.

Tinche:
Thanks for your effort. I will do my best to review over the weekend.

tests/test_multistrategy_dispatch.py
```diff
@@ -0,0 +1,50 @@
...
+def _foo_func():
+    pass
+
+
+def _foo_cls():
+    pass
+
+
+def test_multistrategy_dispatch_register_cls():
+    _fallback()
```
toumorokoshi: If I didn't add these calls, the coverage wasn't sufficient. 

I'm not sure how to tackle that, so I just call them.

tests/test_structure.py
```diff
@@ -279,3 +301,23 @@ def test_structuring_unsupported(converter):
         converter.structure(1, Converter)
     with raises(ValueError):
         converter.structure(1, Union[int, unicode])
+
+
+def test_subclass_registration_is_honored(converter):
```
toumorokoshi: A unit test for the behavior you brought up in the PR.

cattr/multistrategy_dispatch.py
```diff
@@ -0,0 +1,53 @@
...
+    """ a dummy object to help signify a dispatch not found """
+    pass
+
+
+class MultiStrategyDispatch(object):
+    """
+    MultiStrategyDispatch uses a
+    combination of FunctionDispatch and singledispatch.
+
+    singledispatch is attempted first. If nothing is
```
toumorokoshi: Just wanted to note here: we discussed using FunctionDispatch, then falling back to SingleDispatch, but I think that won't work to resolve subclasses: registering a subclass of an iterable, then having the functiondispatch pick that up when determined as a subclass for typing.List and runs iterable_seq.

So switched the order: singledispatch is attempted, then FunctionDispatch as a fallback.

Tinche:
Looking good! Almost ready for merging.

Some comments/questions:
* Can we remove the vendored typing for Python 2?
* Some classes should be converted to attrs classes :) (FunctionDispatch and _DispatchNotFound for sure.)
* Please put in a line in HISTORY.rst about the change.

toumorokoshi:
Ok, addressed 2 and 3.

For 1: I didn't want to pollute this PR with a bunch of changes on the import statements, so I created another pr (#12).

Tinche:
Thank you for your contributions!

Just FYI, I'll probably change the API for registering hooks a little before release (by adding a special call for generics). Also I don't plan on releasing until attrs 17.3 is out because then we can remove cattr.typed in favor of attr.ib(type=).

toumorokoshi:
yep, both of those makes sense. I'd still like to have some way to register a function hook publicly though: that will enable me to register non-class based systems like schematics.

Thanks!

PR #12, opened 2017-10-04
Removing vendored typing

toumorokoshi:
This removes the vendored typing module, for all versions of Python. 

Since singledispatch is not relied upon for the typing modules, the subclass checks are properly supported regardless of the version. 

This should be merged in after #11


Tinche:
Please rebase this on master :)

Tinche:
Thanks!

PR #18, opened 2017-12-25
Some easy optimizations

bluetech:
Hi,

I started using cattrs, but noticed it was slower than my previous hand-rolled code, so I added a micro-benchmark and performed some easy optimizations based on it. Please see the commit messages for details.

There are more optimizations possible of course, hopefully someone else will try :)

Also, I only looked at the dict-based code. Tuples are faster, but I use dicts for extensibility.

### DIFF
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -1,0 +1,187 @@
+from typing import Optional, List
+import enum
+import attr
+import cattr
+import cProfile
+
+
+@attr.s(slots=True, frozen=True)
+class Lorem:
+    ipsum = attr.ib()
+    dolor = attr.ib()
+    sit = attr.ib()
+    amet = attr.ib()
+    consectetur = attr.ib()
+    adipiscing = attr.ib()
+
+
+@attr.s(slots=True, frozen=True)
+class Fugiat:
+    eiusmod = attr.ib()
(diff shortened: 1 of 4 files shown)


bluetech:
Pushed a couple more commits. Travis fails because of hypothesis `FailedHealthCheck` which seems unrelated. Locally `tox` passes.

Tinche:
Hello,

I definitely agree the next step for cattrs is working on making it very fast. I'll take a closer look in the next couple of days.

Tinche:
I'll gladly take the benchmark example and the inlining, but I have reservations about the LRU caches. I have plans on making cattrs faster using a slightly different approach and I worry about adding too many caches at different layers. So if you could revert those I can merge this in.

I'm not saying the caches are a bad idea, I'm saying let's take this one step at a time. :)

bluetech:
> I worry about adding too many caches at different layers.

Just to be clear - these lru caches replace already existing caches (based on dicts), they are not new caches. It's just that the lru caches are faster. Should I still remove them? (No problem if so).

(Also, I pushed another commit with a tiny simplification I just noticed).

Tinche:
You're right, let's get this in and leave future tweaks for the future.

Thanks for your contribution :)

PR #28, opened 2018-02-17
Ensure that LICENSE is included in the built wheel

pombredanne:
Without this directive it is not included: wheels ignore MANIFEST.in

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


Tinche:
Thanks!

Issue #13, opened 2017-10-13
Self-referential type specification issues


altendky:
* cattrs version: 02d9da0a0a7a81c66fc9d8c56c2457170333bd7b
* Python version: 3.6.2
* Operating System: Kubuntu 16.04

### Description

Proper definition of self-referential types seems unclear.  This would relate to [the forward declarations problem discussed in PEP484](https://www.python.org/dev/peps/pep-0484/#the-problem-of-forward-declarations).  I started digging on this because of the `typing.List[typing.Union['WithUnion', WithChild]]` case but provide a few attempts building up to that for reference.

My big picture is to easily [de]serialize attrs classes which I am using for a PyQt5 tree model.  They can allow multiple types of children within lists (a group node would be able to hold both other groups and parameters, for example).

In some of the cases self-reference seems achievable by registering `typing._ForwardRef('MyClass')` but  it doesn't seem clean to have to register a private member of `typing` so I thought it would be worth considering a cleaner approach.  I haven't worked with `typing` before so I'm not sure exactly how it goes about dealing with class name strings and disambiguation of matching class names in different scopes or modules.

As a side note, I first rolled my own [de]serialization scheme around `attrs`.  Then started doing an `attrs`/`marshmallow` integration in [`graham`](https://github.com/altendky/graham).  Now I am trying out `cattrs` for the same roll.

### What I Did

```
import traceback
import typing

import attr
import cattr


@attr.s
class WithChild:
    child = cattr.typed(
        default=None,
        type='WithChild',
    )


@attr.s
class WithChildren:
    children = cattr.typed(
        default=attr.Factory(list),
        type=typing.List['WithChildren'],
    )


@attr.s
class WithUnion:
    children = cattr.typed(
        default=attr.Factory(list),
        type=typing.List[typing.Union['WithUnion']]
    )


@attr.s
class WithUnionTwo:
    children = cattr.typed(
        default=attr.Factory(list),
        type=typing.List[typing.Union['WithUnion', WithChild]]
    )


def tryit():
    try:
        cattr.structure({}, WithChild)
    except:
        print(' - - - - - - WithChild failed')
        print(traceback.format_exc())
    else:
        print(' - - - - - - WithChild succeeded')

    try:
        cattr.structure({'children': [{}]}, WithChildren)
    except:
        print(' - - - - - - WithChildren failed')
        print(traceback.format_exc())
    else:
        print(' - - - - - - WithChildren succeeded')

    try:
[...]

PR #42, opened 2018-09-15
Add support for typing.ForwardDef

vodik:
Its also correct for type annotations to be strings that are evaluated at run time. Internally, Python converts them into typing.ForwardDef
objects and evals them at runtime. cattr needs to do the same thing as
well.

This also brings in compatibility for PEP 563 where deferred
evaluation becomes the default behaviour.

### DIFF
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -9,6 +9,11 @@ from typing import (
     Tuple,
 )
 
+try:
+    from typing import ForwardRef
+except ImportError:
+    from typing import _ForwardRef as ForwardRef  # noqa
+
 version_info = sys.version_info[0:3]
 is_py2 = version_info[0] == 2
 is_py3 = version_info[0] == 3
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -143,13 +143,14 @@ class Converter(object):
         self._unstructure_func.register_cls_list([(cls, func)])
 
     def register_unstructure_hook_func(self, check_func, func):
+        # type: (Callable[Any], Callable[[T], Any]) -> None
         """Register a class-to-primitive converter function for a class, using
         a function to check if it's a match.
         """
-        # type: (Callable[Any], Callable[T], Any]) -> None
         self._unstructure_func.register_func_list([(check_func, func)])
 
     def register_structure_hook(self, cl, func):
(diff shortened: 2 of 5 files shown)


vodik:
Yeah, so this is a change of the public API, but I don't think it can be fixed without doing that as we need to evaluate the annotations in the right namespace and currently that information isn't preserved.

This should provide a solution to both #42 and #13 (well, it will for #13, but I haven't actually fixed up the `Union` structurer just yet).

vodik:
I've just noticed on attrs [#265](https://github.com/python-attrs/attrs/issues/265) that they added an `attr.resolve_types`.

I'm not sure if that function, or `get_type_hints` is sufficient to implement this properly as its possible to eval out a `ForwardRef`'d type, which would then still have to be manually eval'ed.

Tinche:
Hi, thanks for looking into this. This is definitely something we'll need to handle soon. I'm not sure what the best solution is, but I suspect it's actually generating a class specific structure/unstructure function once and caching it. Then, at generation time, the types can be resolved.

j-walker23:
@Tinche First, this library is awesome. Thanks for it.
Just wanted to ask if any help on this was needed or have you just gotten busy with other things. No rush of course, just curious. Would love to take advantage of this super cool lib!

johnwlockwood:
This fails on:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from typing import List, Optional
from cattr import structure, unstructure
import attr

@attr.s
class Boop(object):
    value: int = attr.ib()
    boop: Optional['Boop'] = attr.ib(default=None)


def main():
    b = {
        'value': 1,
        'boop': {
            'value': 2,
            'boop': {
                'value': 3,
                'boop': None,
            }
        }
    }
    boops = structure(b, Boop)
    print(boops)


if __name__ == "__main__":
    main()
```
`ValueError: Unsupported type: ForwardRef('Boop'). Register a structure hook for it.`

jacobg:
Any update on resolving this issue?

Tinche:
This is obsolete if you use from __future__ import annotations (i.e. deferred annotation processing as per PEP563), which we support, so closing this.

Tinche:
Note: this is supported using `cattr.GenConverter`, which will become the default in one of the next versions.

Issue #23, opened 2017-12-31
Attrs and underscore attributes


adrientetar:
So attrs has this feature where underscore-prefixed attributes are expected without the underscore in the ctor. But that means `cattr.structure(cattr.unstructure(myinstance), myclass)` will traceback with unexpected keyword argument.

Since cattrs is specifically made to work with attrs, shouldn't it roundtrip underscore-prefixed attributes by default?

atugushev:
@Tinche do you have any idea how to workaround this case? See also #89.

Tinche:
Can you clarify the issue? cattrs works with private attrs attributes, and has for a long time (the OP is from 2017).

atugushev:
@Tinche 

The issue copy-pasted from https://github.com/Tinche/cattrs/issues/89#issue-636023895:

```python
# t.py

import attr
import cattr


@attr.s(auto_attribs=True)
class foo:
    a: str = attr.ib()
    _b: str = attr.ib()


foo(a="a", b="b")
# foo(a='a', _b='b')

cattr.unstructure(foo(a="a", b="b"))
# {'a': 'a', '_b': 'b'}

# This somehow works despite the init for foo() looking for `b` not `_b`
cattr.structure({"a": "a", "_b": "b"}, foo)
# foo(a='a', _b='b')

# But this fails because cattr fails to pass `b` into the __init__ function
cattr.structure({"a": "a2", "b": "b2"}, foo)
```

Traceback is:

```console
$ python t.py
Traceback (most recent call last):
  File "t.py", line 22, in <module>
    cattr.structure({"a": "a2", "b": "b2"}, foo)
  File "/private/var/folders/4x/rz_w89dx1z3bydfd6qpdjn_h0000gn/T/tmp.QVcREd50/.venv/lib/python3.8/site-packages/cattr/converters.py", line 199, in structure
    return self._structure_func.dispatch(cl)(obj, cl)
  File "/private/var/folders/4x/rz_w89dx1z3bydfd6qpdjn_h0000gn/T/tmp.QVcREd50/.venv/lib/python3.8/site-packages/cattr/converters.py", line 339, in structure_attrs_fromdict
    return cl(**conv_obj)  # type: ignore
TypeError: __init__() missing 1 required positional argument: 'b'
```

Tinche:
Ah, you want the unstructured data to not match the class? You can customize using the new `GenConverter`.

```
import attr

import cattr
[...]

PR #45, opened 2018-10-12
strip leading underscores from attributes so can be structured.

wakemaster39:
resolves issue #23

### DIFF
--- a/docs/structuring.rst
+++ b/docs/structuring.rst
@@ -326,7 +326,7 @@ annotations when using Python 3.6+, or by passing the appropriate type to
     ...     a: int = attr.ib()
     ...
     >>> attr.fields(A).a
-    Attribute(name='a', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, metadata=mappingproxy({}), type=<class 'int'>, converter=None)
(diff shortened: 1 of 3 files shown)


anthrotype:
@Tinche what do you think of this?
It’s the single most annoying thing for me when working with attrs/cattrs. I often use the leading underscore for read-only attributes that are accessible via `@property` getters.

Tinche:
Agreed, both with the original submission and your suggestion. @wakemaster39 if you can apply @anthrotype 's suggestion I will merge this in :)

wakemaster39:
I made the change, I personally prefer startswith over [0] for readability but I am also not typically involved in performance in python.

wakemaster39:
i did my best to make all the test pass, but I figured not adding in a # noqa for the linting error would be fine. Also I have never worked with hypothesis before so I am not sure how to add a test properly for _attribute_name.

 I'll read up on it, but I don't know if I will be able to sort it out in a reasonable time span.

Tinche:
Thanks for this!

DanielJoyce:
I'm still seeing this failure. Did this merge fail or never happen?

DanielJoyce:
Looks like the merge failed and never went through.

Issue #53, opened 2018-12-05
cattr.structure converting None


fwannmacher:
* cattrs version: 0.9.0
* Python version: 3.6
* Operating System: Ubuntu 18.04

### Description

structure is converting None to 'None' when it is set to a str field.

### What I Did

```
dataclass = partial(attrs, auto_attribs=True, frozen=True)

@dataclass
class BaseData:
    def to_dict(self):
        return asdict(self)

@dataclass
class A(BaseData):
    a: str
    b: str

print(structure({'a': 'asd', 'b': None}, A))
>>> A(a='asd', b='None')
```

madsmtm:
I can reproduce in `0.9.1.dev0`, the latest `master`, using a simpler example:
```py
>>> import attr, cattr
>>> @attr.s
>>> class Model:
...     a: str = attr.ib()
...
>>> cattr.structure({"a": None}, Model)
Model(a='None')
```
The issue is related to #26, and arise due to the type coercion of `str`

Tinche:
Going to close this as wontfix.

The solution here is to define your own structure hook for `str` that does an isinstance check instead of calling `str` on the value.

See https://catt.rs/en/stable/defaulthooks.html#primitive-values for more details.

Issue #26, opened 2018-01-18
Type validation on structuring?


markrwilliams:
* cattrs version: 0.6.0, [master](https://github.com/Tinche/cattrs/tree/0d1c3b3912f59fcea7cfd6d3ea5238810cb7f273)
* Python version: 2.7.14, 3.6.4
* Operating System: Linux

### Description

`cattrs.unstructure` implicitly converts dictionary members to the type of the target class' `Attribute`.  This means I can't use `cattrs` to enforce basic type constraints on deserialized data.  Is it a design decision?

### What I Did

The following program allows a list to be deserialized as a string and a string as an integer:

```python
import attr
import cattr


@attr.s
class Wat(object):
    a_string = attr.ib(type=str)
    an_integer = attr.ib(type=int)


print(cattr.structure({"a_string": [1, 2, 3], "an_integer": "10"}, Wat))
```

This prints `Wat(a_string='[1, 2, 3]', an_integer=10)`.

`cattrs` silently converts the list provided as `a_string` to a `str` because [`Converter._structure_unicode` calls `str` on an object if it isn't `bytes`/`str` or `str`/`unicode`](https://github.com/Tinche/cattrs/blob/0d1c3b3912f59fcea7cfd6d3ea5238810cb7f273/cattr/converters.py#L226-L231).  

The string provided as `an_integer` becomes an integer because of a more general problem: [ primitive types other than strings are structured with `Converter._structure_call`](https://github.com/Tinche/cattrs/blob/0d1c3b3912f59fcea7cfd6d3ea5238810cb7f273/cattr/converters.py#L93-L96), which in turn [directly applies the specified type's constructor to the received object](https://github.com/Tinche/cattrs/blob/0d1c3b3912f59fcea7cfd6d3ea5238810cb7f273/cattr/converters.py#L217-L224).

Raising an exception when a value doesn't match the expected type is a valuable tool in writing robust deserialization code.  Right now, I have to use something like [schema](https://github.com/keleshev/schema) with `cattrs` to achieve this, which unfortunately requires that I write my class' type specifications twice.

It's not possible to work around this with a `attrs` validator (e.g., with `a_string = attr.ib(type=str, validator=attr.validators.instance_of(str))`) because by the time `Wat`'s initializer the value has been already been converted to the target type.  It _is_ possible to use a custom `type` wrapper, but this requires registering structuring hooks and wreaks havoc with `mypy`.

[...]

PR #54, opened 2019-01-24
Allow renaming of properties with attrs

sebastianmika:
There are scenarios where the unstructured data have properties that are forbidden keywords in python. In an `attrs` class this property needs to take a different name.

Example data: `{'class': 'not allowed'}`

The following `attrs` class is **invalid** - at least I wouldn't see how that can be made working in `attrs` nor `cattr`:

```python
@attr.s
class MyClass(object):
    class = cattr.ib(type=str)
```

Proposal: use `attrs` metadata and provide an extension to `cattr` that allows to remap the invalid name to something valid when structuring and vice-versa when unstructuring.

I added a test, - but the `hypothesis` business is beyond my skills - sorry.

This works for me and is a solution similar to that taken by other packages (e.g. `related`):

```python
import attr
import cattr


@attr.s
class MyClass(object):
    renamed = cattr.ib(src_key='class', type=str)


def test_structure_keywords():
    data = {'class': 'value'}
    obj = cattr.structure(data, MyClass)
    assert cattr.unstructure(obj) == data
```

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -204,8 +204,9 @@ class Converter(object):
         rv = self._dict_factory()
         for a in attrs:
             name = a.name
+            from_key = a.metadata.get(CATTRS_METADATA_KEY, a.name)
             v = getattr(obj, name)
-            rv[name] = dispatch(v.__class__)(v)
+            rv[from_key] = dispatch(v.__class__)(v)
         return rv
 
     def unstructure_attrs_astuple(self, obj):
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -27,11 +27,15 @@ from ._compat import (
 )
 from .disambiguators import create_uniq_field_dis_func
 from .multistrategy_dispatch import MultiStrategyDispatch
+import attr
 
 
 NoneType = type(None)
 T = TypeVar("T")
 V = TypeVar("V")
+
+
+CATTRS_METADATA_KEY = 'cattrs_structure_key'
 
 
 class UnstructureStrategy(Enum):
@@ -301,9 +305,10 @@ class Converter(object):
             # We detect the type by metadata.
             type_ = a.type
             name = a.name
+            from_key = a.metadata.get(CATTRS_METADATA_KEY, a.name)
 
             try:
-                val = obj[name]
+                val = obj[from_key]
             except KeyError:
                 continue
 
@@ -433,3 +438,11 @@ class Converter(object):
                 "currently. Register a loads hook manually."
             )
         return create_uniq_field_dis_func(*union_types)
+
+    def ib(self, default=attr.NOTHING, validator=None, repr=True, cmp=True,
+           hash=None, init=True, convert=None, metadata={}, src_key=None):
+        metadata = dict() if not metadata else metadata
+        if src_key:
+            metadata[CATTRS_METADATA_KEY] = src_key
+        return attr.ib(default, validator, repr, cmp, hash,
+                       init, convert, metadata)
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -204,9 +204,10 @@ class Converter(object):
         rv = self._dict_factory()
         for a in attrs:
             name = a.name
-            from_key = a.metadata.get(CATTRS_METADATA_KEY, a.name)
+            # get src_key if any and unstrucutre to that instead of a.name
+            src_key = a.metadata.get(CATTRS_METADATA_KEY, a.name)
             v = getattr(obj, name)
-            rv[from_key] = dispatch(v.__class__)(v)
+            rv[src_key] = dispatch(v.__class__)(v)
         return rv
 
     def unstructure_attrs_astuple(self, obj):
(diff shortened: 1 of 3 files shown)


src/cattr/converters.py
```diff
@@ -433,3 +441,21 @@ def _get_dis_func(self, union):
                 "currently. Register a loads hook manually."
             )
         return create_uniq_field_dis_func(*union_types)
+
+    def ib(self, default=attr.NOTHING, validator=None, repr=True, cmp=True,
+           hash=None, init=True, convert=None, metadata=None, type=None,
+           converter=None, factory=None, kw_only=False, src_key=None):
```
madsmtm: Instead of the name `src_key`, we could also use something else. `Related` uses [`key`](https://github.com/genomoncology/related/blob/master/tests/ex05_field_names/test_renamed.py). `Marshmallow` uses [`data_key`](https://marshmallow.readthedocs.io/en/3.0/quickstart.html#specifying-serialization-deserialization-keys).

madsmtm:
I don't think you have to worry about the failing tests, they're due to a change in `hypothesis` [`v4.0.0`](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-0-0)

src/cattr/converters.py
```diff
@@ -27,13 +27,17 @@
...
 from .multistrategy_dispatch import MultiStrategyDispatch
+import attr
 
 
 NoneType = type(None)
 T = TypeVar("T")
 V = TypeVar("V")
 
 
+CATTRS_METADATA_KEY = 'cattrs_structure_key'
```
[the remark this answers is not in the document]
sebastianmika: I tried to follow the advise made here: http://www.attrs.org/en/stable/extending.html. I agree that it looks weird in my code to have an `cattr.ib`.

An alternative would be to use a more distinctive name like in the referenced example, e.g. `cattr.ib_mapped(..., src_key='mapped')`.

Of course, also this would work

```python
from cattr import CATTRS_METADATA_KEY

attrs.ib(..., metadata={CATTRS_METADATA_KEY: 'mapped'})
```
But it looks very ugly - at least `CATTRS_METADATA_KEY` should be more self explanatory + shorter.

What do you think @Tinche? Being active on `attrs` itself you certainly have a strong opinion.

src/cattr/converters.py
sebastianmika: I am absolutely agnostic about the name. Maybe `key` would actually be a better choice as we are mapping from a dictionary `key` to the property and back to the `key`.

I'll change it to be that.

src/cattr/converters.py
```diff
@@ -318,6 +318,10 @@ def structure_attrs_fromdict(self, obj, cl):
             if name[0] == "_":
                 name = name[1:]
 
+            if val is None and type_ is not None:
+                raise ValueError('property "{:s}" of type "{:s}" must not be None'.format(
+                    src_key, str(type_)))
+
```
madsmtm: Would this fix #53? And would it handle `Optional[str]`? I think it should be a separate PR though, since there's a bit of uncertainty as to what the best approach is, see #26

Tinche:
Hi,

thank you for working on cattrs. I think this is a very useful feature to have, but I would prefer a different direction.

I think renaming attributes should be part of the (de)serialization logic, not the class itself, so I don't really like having it as an attribute in attr.ib(), even though I know other libraries and languages might use this approach (for example, Go).

On the master branch I've pushed an experimental new feature: a way of automatically generating unstructuring functions for attrs classes, and overriding their attributes. Here's an example:

```
import attr
from cattr import Converter
from cattr.generation import make_dict_unstructure_fn, override

@attr.s
class TestClass:
    a: str = attr.ib(default="a")

c = Converter()
c.register_unstructure_hook(TestClass, make_dict_unstructure_fn(TestClass, c, a=override(omit_if_default=True)))

>>> c.unstructure(TestClass())
{}   # 'a' had its default value so it was omitted.
```

We could easily support something like this:
```
@attr.s
class TestClass:
    klass: str = attr.ib(default="a")

c.register_unstructure_hook(TestClass, make_dict_unstructure_fn(TestClass, c, a=override(rename="class")))
```

These generated functions should be significantly faster than generic ones, and can support various overrides. The idea is, down the line, to have the converters generate them automatically, while also being overridable. The API is also clunky, and I have an idea on improving it.

The feature is not done - I still need to finish tests, iron out the implementation (like add support for generating structuring functions too) and write the docs. However feel free to take a look. I think adding rename support would be pretty easy, and since this is serialization logic, this is the right place for it to live.

madsmtm:
To me, speed doesn't really matter, but I like the fact that I could now have two different converters, that could structure a `TestClass` from two different data sources! Also, this would allow us to easily use these features on external `attrs` classes. 👍

Just for reference, this is similar to the way `dacite` does it, see [dacite#remapping](https://github.com/konradhalas/dacite#remapping)

sebastianmika:
I care a lot for speed - actually speed is a reason I would need to look for some other solution than `cattr`. So: I love you proposal @Tinche and happily close this PR. 

Time permitting (and wrapping my head around what you do there) I will try to contribute my 5 Cents to add the renaming logic.

Issue #60, opened 2019-04-06
add py.typed for MyPy


petergaultney:
* cattrs version: 1.0.0.rc0

### Description

If `cattrs` is going to start supporting MyPy type checking (meaning I don't need my own stubs anymore), I think the standard practice, following PEP 561, is to add an empty `py.typed` file to the top level of the package alongside `__init__.py`.  This lets MyPy know that this package is safe to type-check. 

Maybe 1.0.0 isn't fully there, in which case it probably isn't time for this. But just in case it had been overlooked...

Tinche:
Thanks, this is definitely something I'd be interested in. Consider submitting a quick PR?

petergaultney:
PR made! I would also be happy to spend some additional time working on the internal typing of cattrs, if you're interested. I use `cattrs` pretty heavily at my job and am invested in the success of the project. :)

PR #62, opened 2019-05-03
added py.typed to src and distribution

petergaultney:
Additionally made a small improvement to the type signature of a function in `generation.py`

Addresses #60


Tinche:
Thanks!

PR #66, opened 2019-08-30
Fix master build

joeldodge79:
Fix test_unmodified_generated_unstructuring test

 This was broken for 3.7 due to how `cattr.generation.make_dict_unstructure_fn`
reaches directly into `converter._unstructure_func.dispatch` to get the
converter function passing the annotated type of the field. In 3.7 the
`TypeError: Subscripted generics cannot be used with class and instance checks`
issue causes the `dispatch` to be the fallback identity converter func.

Normal usage (i.e. unstructure_attrs_asdict) passes in field_inst.__class__
which would never hit the above `TypeError` so this issue is unique to
`make_dict_unstructure_fn`

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -68,7 +68,7 @@ Ready to contribute? Here's how to set up `cattrs` for local development.
 
     $ mkvirtualenv cattrs
     $ cd cattrs/
-    $ pip install -e .
+    $ pip install -e .\[dev\]
 
 4. Create a branch for local development::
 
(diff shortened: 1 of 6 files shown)


Tinche:
Thanks, will take a look over the weekend!

joeldodge79:
I got the same `DeadlineExceeded` for pypy locally but was hoping it would somehow "just run faster" on travis... Apparently not :-) Locally I configured the "CI" profile settings with `deadline=None` to get pypy to pass. Not sure what the best approach is here. For now I'll add that change to see if the build at least passes

Tinche:
Thank you!

Issue #67, opened 2019-09-04
attrbute err with attr.ib(init=False)


swuecho:
* cattrs version: 0.9.0
* Python version: 3.7
* Operating System: mac

### What I Did

```

import attr, cattr

@attr.s(cmp=False)
class C:
    a = attr.ib()
    b = attr.ib(init=False)

if __name__ == "__main__":
    instance = C(1)
    print(cattr.unstructure(instance))
```
will throw err

```
AttributeError: 'C' object has no attribute 'b'
```

PR #68, opened 2019-09-04
check obj hasattr

swuecho:
when attr.ib(init=False), obj might not has the attr

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -200,8 +200,9 @@ class Converter(object):
         rv = self._dict_factory()
         for a in attrs:
             name = a.name
-            v = getattr(obj, name)
-            rv[name] = dispatch(v.__class__)(v)
+            if hasattr(obj, name):
+                v = getattr(obj, name)
+                rv[name] = dispatch(v.__class__)(v)
         return rv
 
     def unstructure_attrs_astuple(self, obj):


swuecho:
#67

sveinse:
What is the status on this? This fix is something I need.

swuecho:
> What is the status on this? This fix is something I need.

I use my own fork as a workaround.

add `git+git://github.com/swuecho/cattrs.git@patch-1#egg=cattrs` to requirements.txt

anthrotype:
I agree init=False attributes should not be unstructured.
@Tinche can you please take a look at this?

swuecho:
any update? the change seems simple enough and pass all the tests.

src/cattr/converters.py
```diff
@@ -194,8 +194,9 @@ def unstructure_attrs_asdict(self, obj):
         rv = self._dict_factory()
         for a in attrs:
             name = a.name
-            v = getattr(obj, name)
-            rv[name] = dispatch(v.__class__)(v)
+            if hasattr(obj, name):
+                v = getattr(obj, name)
+                rv[name] = dispatch(v.__class__)(v)
```
tmewett, line 199: This will only unstructure attributes that are set on the instance. This works when the `init=False` attributes are not set at unstructure time, but how does it work if they are? If the attribute is set, won't the structuring code attempt to initialise it? (See `structure_attrs_fromdict`.) I think we need to explicitly check if an attribute has `init` set to true, and exclude ones that don't, whether they are actually set on the instance or not.

Additionally, this behaviour needs to be added to `unstructure_attrs_astuple` below.

tmewett:
We ought to add a test for #67 when it is solved.

src/cattr/converters.py
swuecho: The code only deal with un-structure, if the attribute is set, the attribute will be unstructured, which is the default behavior.

swuecho:
#93 for update

PR #69, opened 2019-09-11
Add support for typing.ForwardRef

prOOrc:
Add support for typing._ForwardDef without break backward compatibility


JasonGilholme:
Hi - this looks really handy and we have a current need for it.  Any plans to merge this in soon?

anthrotype:
I'd love to see support for typing.ForwardRef in cattrs!

wakemaster39:
is it worth adding this in like this? `attrib(type=XXXX)` is only available for backwards compatibility and the [recommended](http://www.attrs.org/en/stable/api.html#attr.ib) way is using type annotations `x: str`.

The change to handle the `annotations` future is then much smaller in you need to only a slight modification to the structure attrs functions:

```
        type_hints = get_type_hints(cl)
        for a in cl.__attrs_attrs__:  # type: ignore
            if not a.init:
                continue
            # We detect the type by metadata.
            name = a.name
            type_ = type_hints[name]
```

I have been playing with this to make sure it works in 3.8 under all conditions I have run into, and it seems to solve the problem so far.

anthrotype:
@wakemaster39 I like the simplicity of using `get_type_hints` to resolve ForwardRefs. However that would only work for auto_attribs=True kinds of attrs classes. Maybe doing `type_ = type_hints.get(name, a.type)` one could support structuring both old attrs classes with `type` metadata and new ones with annotations.

ex-nerd:
Given more activity on this PR, is this more likely to be the preferred solution compared to the one submitted as #42?

johnwlockwood:
This failed on:
```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from typing import List, Optional
from cattr import structure, unstructure
import attr


@attr.s
class Node(object):
    value: int = attr.ib()
    children: Optional[List['Node']] = attr.ib(default=None)


def main():
    g = {
        'value': 1,
        'children': [
            {'value': 3},
            {'value': 4, 'children': [{'value':44}]},
            {'value': 5, 'children': [{'value':55}]}
        ]
    }
    n = structure(g, Node)
    print(n)


if __name__ == "__main__":
    main()
```

```zsh
  File "recursive_struct.py", line 29, in <module>
    main()
  File "recursive_struct.py", line 24, in main
    n = structure(g, Node)
  File "/Users/john/.pyenv/versions/anaconda3-2019.10/lib/python3.7/site-packages/cattr/converters.py", line 204, in structure
    return self._structure_func.dispatch(cl)(obj, cl)
  File "/Users/john/.pyenv/versions/anaconda3-2019.10/lib/python3.7/site-packages/cattr/converters.py", line 327, in structure_attrs_fromdict
    dispatch(type_)(val, type_) if type_ is not None else val
  File "/Users/john/.pyenv/versions/anaconda3-2019.10/lib/python3.7/site-packages/cattr/converters.py", line 400, in _structure_union
    return self._structure_func.dispatch(other)(obj, other)
  File "/Users/john/.pyenv/versions/anaconda3-2019.10/lib/python3.7/site-packages/cattr/converters.py", line 340, in _structure_list
    for e in obj
  File "/Users/john/.pyenv/versions/anaconda3-2019.10/lib/python3.7/site-packages/cattr/converters.py", line 340, in <listcomp>
    for e in obj
  File "/Users/john/.pyenv/versions/anaconda3-2019.10/lib/python3.7/site-packages/cattr/converters.py", line 415, in _structure_forward_ref
    owner_class = self._forward_ref_owners_registry[ref]
KeyError: ForwardRef('Node')
```

Tinche:
This is obsolete if you use `from __future__ import annotations` (i.e. deferred annotation processing as per PEP563), which we support, so closing this.

PR #73, opened 2019-10-27
Add support for python 3.8

mikelane:
Adds support for python 3.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ matrix:
       env: TOXENV=py36
     - python: "3.7"
       env: TOXENV=py37
+    - python: "3.8"
+      env: TOXENV=py38
       dist: xenial
       sudo: true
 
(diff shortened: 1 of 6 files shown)


Tinche:
Thanks!

avanov:
Has this been released on PyPy? The currently available pre-release still fails on 3.8

causey:
I'm still seeing this import issue; when is it planned to be released?

DiscreteTom:
Installing the master version from github by running `pip install git+https://github.com/Tinche/cattrs.git` can temporary fix this before the latest version is released on PyPI.

DrLuke:
Thanks for the PR!

However, with the release of python 3.9 this approach will break again?

mikelane:
> Thanks for the PR!
> 
> However, with the release of python 3.9 this approach will break again?

Yes, this will break in the same way for python 3.9, but I wanted to be conservative with my approach.

Tinche:
I released 1.0.0 a few days ago with 3.8 support fwiw. We've upgraded at work to 3.8 and now I see we're a tiny bit buggy on 3.8, so a 1.0.1 release is probably incoming.

PR #75, opened 2019-11-01
Update to be compatible with attrs 19.3.0 (only)

sebastianmika:


sebastianmika:
Sorry - did mean to create the PR here

PR #76, opened 2019-11-01
py3.8 support

IlyaSkriblovsky:
Hey, thanks for great work!

`>=` checking for 3.7 seems more reasonable than strict `==`

cattrs fail to import on py3.8 otherwise


IlyaSkriblovsky:
Sorry, #73 seems better

PR #85, opened 2020-04-26
Fixes #41 pep 563

### LINKED ISSUE #41: Breaks when PEP 563 is active
* cattrs version: All versions, as far as I can tell
* Python version: Python 3.7
* Operating System: Arch Linux

### Description

`cattr.structure` breaks when [PEP 563](https://www.python.org/dev/peps/pep-0563/) is activated. This can be done in Python 3.7 with the new `from __future__ import annotations`.

In short, this defers the evaluation of type annotations. So consider this class:

```python
@attr.s(auto_attribs=True)
class Foo:
    bar: str
```

Traditionally, `__annotations__` looks like this: 

```
>>> Foo.__annotations__
... {'bar': str}
```

but under PEP 563 (and eventually the default behaviour), it'll look like this:

```
>>> Foo.__annotations__
... {'bar': 'str'}
```

This seems to confuse cattr and causes problems. Round-tripping the above class, we now see exceptions like this:

```ipython
In [16]: cattr.structure(cattr.unstructure(Foo('Hello World')), Foo)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-16-1bbcc538d3df> in <module>()
----> 1 cattr.structure(cattr.unstructure(Foo('Hello World')), Foo)

~/.local/share/virtualenvs/aastrology-5jK3bc4O/lib/python3.7/site-packages/cattr/converters.py in structure(self, obj, cl)
    176         """Convert unstructured Python data structures to structured data."""
    177         # type: (Any, Type) -> Any
--> 178         return self._structure_func.dispatch(cl)(obj, cl)
    179 
    180     # Classes to Python primitives.

~/.local/share/virtualenvs/aastrology-5jK3bc4O/lib/python3.7/site-packages/cattr/converters.py in structure_attrs_fromdict(self, obj, cl)
    294             except KeyError:
    295                 continue
--> 296             conv_obj[name] = dispatch(type_)(val, type_)
    297 
    298         return cl(**conv_obj)

~/.local/share/virtualenvs/aastrology-5jK3bc4O/lib/python3.7/site-packages/cattr/converters.py in _structure_default(self, obj, cl)
    238             "it.".format(cl)
    239         )
--> 240         raise ValueError(msg)
    241 
    242     def _structure_call(self, obj, cl):

ValueError: Unsupported type: str. Register a structure hook for it.
```

This fix is probably to apply this logic: https://www.python.org/dev/peps/pep-0563/#resolving-type-hints-at-runtime

Here:

https://github.com/Tinche/cattrs/blob/6e90fd28af8e23c44798b5cbd98e8d5a06eb87ba/src/cattr/converters.py#L287-L296

Which I'm currently digging into.

JiveHelix:
When PEP 563 is active, all type information is stored using strings. The actual type can be retrieved by looking up the __module__ where the class was defined, or from __builtins__. I added a couple of functions to converters.py which need the class for the __module__ lookup. A change to _structure_attr_from_tuple was necessary to pass the class through. I added tests/test_563.py, and CI passes for Python 3.7 and 3.8, but fail for 3.5 and 3.6 because of the 'from __future__ import annotations' line in the test files. I tried to remove that with a conditional import, but that line is required to come first. I trust that there is a way to exclude the tests for this fix for 3.6 and earlier.

### DIFF
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -88,6 +88,7 @@ if is_py37 or is_py38:
         else:
             return type_
 
+
 else:
     from typing import _Union
 
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -13,14 +13,13 @@ from typing import (
 version_info = sys.version_info[0:3]
 is_py37 = version_info[:2] == (3, 7)
 is_py38 = version_info[:2] == (3, 8)
-is_py37_or_later = version_info[:2] >= (3, 7)
 
 
 unicode = str
 bytes = bytes
 
 if is_py37 or is_py38:
-    from typing import List, Union, _GenericAlias
+    from typing import List, Union, _GenericAlias, ForwardRef, Any
 
     def is_union_type(obj):
         return (
@@ -74,6 +73,20 @@ if is_py37 or is_py38:
             or args == bare_mutable_seq_args
         )
 
+    @lru_cache(32)
+    def _get_class_namespace(class_: type) -> Dict[str, Any]:
+        return vars(sys.modules[class_.__module__])
+
+    @lru_cache(32)
+    def get_type(parent_class: type, type_: Union[type, str]) -> type:
+        """Returns the actual type if type_ is a string."""
+
+        if isinstance(type_, str):
+            globalns = _get_class_namespace(parent_class)
+            forwardRef = ForwardRef(type_, is_argument=False)
+            return forwardRef._evaluate(globalns, None)
+        else:
+            return type_
 
 else:
     from typing import _Union
@@ -102,3 +115,6 @@ else:
         return not type.__args__
 
     is_bare_frozenset = is_bare
+
+    def get_type(ignored: type, type_: type) -> type:
+        return type_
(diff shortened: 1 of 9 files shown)


JiveHelix:
> When PEP 563 is active, all type information is stored using strings. The actual type can be retrieved by looking up the **module** where the class was defined, or from **builtins**. I added a couple of functions to converters.py which need the class for the **module** lookup. A change to _structure_attr_from_tuple was necessary to pass the class through. I added tests/test_563.py, and CI passes for Python 3.7 and 3.8, but fail for 3.5 and 3.6 because of the 'from **future** import annotations' line in the test files. I tried to remove that with a conditional import, but that line is required to come first. I trust that there is a way to exclude the tests for this fix for 3.6 and earlier.

I was able to adjust the tox.ini file and conftests.py to skip the incompatible tests in py35 and py36.

Darrenmeehan:
Hi @JiveHelix there was some discussion based on your work over in my PR at #88.

Tinche:
This should be fixed when using the new GenConverter (which will become the default in a future release).

PR #86, opened 2020-05-04
Use pycon highlighting in README

adamchainz:
This is the 'python console' lexer in pygments, which is intended for use with `>>>` REPL sessions. It should look better on PyPI on next release.


Tinche:
Thanks!

PR #88, opened 2020-05-20
Add Python 3.9 support

Darrenmeehan:
This is similar to the PR that added Python3.8 support in https://github.com/Tinche/cattrs/pull/73

I've cleaned up the logic somewhat as now the only outlier is python3.6 - so this is the only version we check for.

I've also added `tox` and `travis` config for Python 3.9.

Happy to change anything if required.

## Other Information

I have a `Pipfile` & `Pipfile.lock` locally as I had to install `pytest` & `hypothesis`. Is this something you would like added to the project? Possibly using a `requirements.txt`?

### DIFF
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -11,14 +11,12 @@ from typing import (
 )
 
 version_info = sys.version_info[0:3]
-is_py37 = version_info[:2] == (3, 7)
-is_py38 = version_info[:2] == (3, 8)
-is_py39 = version_info[:2] == (3, 9)
+is_py36 = version_info[:2] == (3, 6)
 
 unicode = str
 bytes = bytes
 
-if is_py37 or is_py38 or is_py39:
+if not is_py36:
     from typing import List, Union, _GenericAlias
 
     def is_union_type(obj):
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -17,7 +17,7 @@ unicode = str
 bytes = bytes
 
 if not is_py36:
-    from typing import List, Union, _GenericAlias
+    from typing import List, Union, _GenericAlias, ForwardRef, Any
 
     def is_union_type(obj):
         return (
@@ -71,6 +71,20 @@ if not is_py36:
             or args == bare_mutable_seq_args
         )
 
+    @lru_cache(32)
+    def _get_class_namespace(class_: type) -> Dict[str, Any]:
+        return vars(sys.modules[class_.__module__])
+
+    @lru_cache(32)
+    def get_type(parent_class: type, type_: Union[type, str]) -> type:
+        """Returns the actual type if type_ is a string."""
+
+        if isinstance(type_, str):
+            globalns = _get_class_namespace(parent_class)
+            forwardRef = ForwardRef(type_, is_argument=False)
+            return forwardRef._evaluate(globalns, None)
+        else:
+            return type_
 
 else:
     from typing import _Union
@@ -99,3 +113,6 @@ else:
         return not type.__args__
 
     is_bare_frozenset = is_bare
+
+    def get_type(ignored: type, type_: type) -> type:
+        return type_
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -13,12 +13,12 @@ from typing import (
 version_info = sys.version_info[0:3]
 is_py37 = version_info[:2] == (3, 7)
 is_py38 = version_info[:2] == (3, 8)
-
+is_py39 = version_info[:2] == (3, 9)
 
 unicode = str
 bytes = bytes
 
-if is_py37 or is_py38:
+if is_py37 or is_py38 or is_py39:
     from typing import List, Union, _GenericAlias
 
     def is_union_type(obj):
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -86,6 +86,7 @@ if not is_py36:
         else:
             return type_
 
+
 else:
     from typing import _Union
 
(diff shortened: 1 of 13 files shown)


Tinche:
This looks good but I will take a closer look soon. I was contemplating dropping support for Pythons below 3.8 for the next version, since 3.8+ supports `Literal[X]` which is potentially a very interesting feature for unions, and to keep complexity down.

First question: what's the performance like with these changes? You can get a feel using pyperf, which is what I use to benchmark. I also wouldn't reduce the lru_cache maxsize parameters, my codebases use a lot of classes :) I also wonder if that cache should be part of the Converter, so it can be overriden.

Second question: where do the classes in the PEP563 examples come from? And why are they exempt from flake8 and black?

Third question: Should we be using `typing.get_type_hints()`?

Darrenmeehan:
Hi @Tinche I should have noted that this PR is now based off https://github.com/Tinche/cattrs/pull/85. I did this in an attempt to get the build passing, as I believe the failures are not due to the changes I've made.

@JiveHelix may be better places to answer these questions.

- A note on dropping Python3.8 - it does seem rather early do be dropping it given it's only recently been released! Did you mean Python3.6 possibly?

Tinche:
@Darrenmeehan Yes, I mean lower than 3.8, so 3.7 and 3.6 namely. Ok, let's switch the discussion over to #85 then.

JiveHelix:
> @Darrenmeehan Yes, I mean lower than 3.8, so 3.7 and 3.6 namely. Ok, let's switch the discussion over to #85 then.

The improvements I added in #85 do not require dropping support for 3.6, it's just that `from __future__ import annotations` is not supported by 3.6, and I added support for type information stored as strings. 3.6 still works as before, there is just no way to test my changes without annotations.

JiveHelix:
> Second question: where do the classes in the PEP563 examples come from? And why are they exempt from flake8 and black?

They are only exempt from flake8 and black when linting with 3.5 and 3.6, because `from __future__ import annotations` is not legal python in those version. They pass flake8 and black for 3.7 and 3.8.

tirkarthi:
`__args__` was removed in https://bugs.python.org/issue40397 and https://bugs.python.org/issue40398 . cattr has a lot of places where __args__ is assumed to be always present. This also results in errors.

```
rg __args__ 
src/cattr/_compat.py
60:    bare_list_args = List.__args__
61:    bare_seq_args = Sequence.__args__
62:    bare_mapping_args = Mapping.__args__
63:    bare_dict_args = Dict.__args__
64:    bare_mutable_seq_args = MutableSequence.__args__
67:        args = type.__args__
101:        return not type.__args__

src/cattr/converters.py
315:        if is_bare(cl) or cl.__args__[0] is Any:
318:            elem_type = cl.__args__[0]
326:        if is_bare(cl) or cl.__args__[0] is Any:
329:            elem_type = cl.__args__[0]
337:        if is_bare(cl) or cl.__args__[0] is Any:
340:            elem_type = cl.__args__[0]
346:        if is_bare(cl) or cl.__args__ == (Any, Any):
349:            key_type, val_type = cl.__args__
369:        union_params = union.__args__
396:        tup_params = tup.__args__
416:        union_types = union.__args__

tests/_compat.py
14:        cl.__args__ = (new_params,)

tests/test_structure.py
109:    if input_set_type.__args__:
110:        set_type = set_type[input_set_type.__args__[0]]
132:    input_set_type.__args__ = (unicode,)
194:    t.__args__ = (t.__args__[0], Optional[t.__args__[1]])
231:    args = t.__args__
235:        and len(args[0].__args__) == 2
236:        and args[0].__args__[1] is NoneType
255:    t.__args__ = args
```

src/cattr/_compat.py
```diff
@@ -73,6 +71,21 @@ def is_bare(type):
             or args == bare_mutable_seq_args
         )
 
+    @lru_cache(32)
+    def _get_class_namespace(class_: type) -> Dict[str, Any]:
+        return vars(sys.modules[class_.__module__])
+
+    @lru_cache(32)
+    def get_type(parent_class: type, type_: Union[type, str]) -> type:
```
wbolster-eiq, line 79: there's not really a reason to avoid `type_` as an argument name; it can be plain `type` as the other functions in this file also do.

also the `Union[type, str]` can probably become `Union[Type, str]` with a capitalized `Type`. this needs a `from typing import Type` at the top of the file.

see https://docs.python.org/3/library/typing.html#typing.Type, especially

> `Type[Any]` is equivalent to `Type` which in turn is equivalent to `type`, which is the root of Python’s metaclass hierarchy.

src/cattr/_compat.py
```diff
@@ -101,3 +114,6 @@ def is_bare(type):
         return not type.__args__
 
     is_bare_frozenset = is_bare
+
+    def get_type(ignored: type, type_: type) -> type:
```
wbolster-eiq, line 118: same here: `type` without underscore suffix as arg name, `Type` (from the `typing` module) for the annotations.

src/cattr/_compat.py
wbolster: the checks done by flake8 (and especially by semi-random plugins) are by their very nature opinionated so that does hardly count as absolute truth.

sure, ideally there would be no clashing names.

however, it is also ‘better to avoid’ building APIs with clunky names (especially when the API is not just for internal use) :wink: i can think of many cases in which arguments (especially keyword arguments) with names that are also built-ins actually _do_ seem the best name. some examples: `type=...`, `min=...`, `max=...`, `filter=...`, `all=...`, `set=...`, `open=...`, `round=...`

Tinche:
Done on my own, but thanks for the effort!

Darrenmeehan:
> Done on my own, but thanks for the effort!

 Sorry it wasn't fit for merge. Great to have the support for 3.9 in.

Tinche:
@Darrenmeehan I went a different way that I had in mind for a while, sorry for the confusion :)

PR #93, opened 2020-08-19
check obj hasattr

swuecho:
fix as_tuple as well and add tests

https://github.com/Tinche/cattrs/issues/67

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -194,15 +194,16 @@ class Converter(object):
         rv = self._dict_factory()
         for a in attrs:
             name = a.name
-            v = getattr(obj, name)
-            rv[name] = dispatch(v.__class__)(v)
+            if hasattr(obj, name):
+                v = getattr(obj, name)
+                rv[name] = dispatch(v.__class__)(v)
         return rv
 
     def unstructure_attrs_astuple(self, obj):
         # type: (Any) -> Tuple
         """Our version of `attrs.astuple`, so we can call back to us."""
         attrs = obj.__class__.__attrs_attrs__
-        return tuple(self.unstructure(getattr(obj, a.name)) for a in attrs)
+        return tuple(self.unstructure(getattr(obj, a.name)) for a in attrs if hasattr(obj, a.name))
 
     def _unstructure_enum(self, obj):
         """Convert an enum to its value."""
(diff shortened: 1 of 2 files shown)


tmewett:
As I said in #68, this doesn't work correctly. Consider the following code

```python
import attr
import cattr

@attr.s
class TestClass:
    a = attr.ib()
    b = attr.ib(init=False)

def test(x):
    print(f"{x}")
    u = cattr.unstructure(x)
    print(f"unstructures to {u}")
    print(f"which structures back to {cattr.structure(u, TestClass)}")

# create an instance without setting b
x = TestClass(1)
test(x)

# now set b and try again
x.b = 2
test(x)
```

This fails the second test on the call to `cattr.structure(u, TestClass)` because `u` has a key "b", however the b attribute of TestClass cannot be initialised as it has init=False.

We need to check whether the attribute has init=True, instead of whether it is set on the object.

swuecho:
> e has init=True, instead of whether it is

The structure part could be done by user instead of library. I understand you want the roundtrip work, and this change only fix the unstructured part.  Welcome anyone who use this feature to compete it.

tmewett:
I apologise if I sounded critical, I do appreciate the time you took to make this PR :) I might base a branch off this myself, and add the structuring side.

Ultimately a decision needs to be made as to how cattrs should handle init=False attrs. We can either

1. exclude them from all unstructuring, so they are not preserved after going through cattrs; or
2. include them (if set) and then set them manually on the instance at the structuring side.

If we do include them, another question is how it should work when using tuples as unstructured form; if we have optional elements in a tuple, I don't think it's possible to disambiguate them, in general. Maybe `attr.NOTHING` could be useful here?

tmewett:
Another complication is classes with attribs with init=False but a default set. In this case it should be possible to (un)structure them unambiguously as tuples. For these reasons, I think not handling init=False attributes at all is the way to go, unless anyone has a solution disambiguating the tuple form?

swuecho:
Appreciate your effort to cover all cases that might happen, but I do not think all the cases are equal. Is there anyone out there unstructure a class as tuple and then want to structure them back? I can not image a case for that.

As I commented before, For `init=False` case. structure can be easily done by user, while unstructure is not.  

My suggestion is to cover the most wanted cases first and deal with others later.

Tinche:
This use case can be handled with our new hook factories (see here: https://cattrs.readthedocs.io/en/latest/unstructuring.html#unstructuring-hook-factories)

PR #97, opened 2020-10-05
Support for Generic Structuring

wakemaster39:
Right now whenever you want to structure a generic it informs you to register a hook, every time you change the type in that generic you need to register a new hook. What this PR addresses is supporting generic structure of attrs classes based on the typing information supplied.

```
from attr import attrs, attrib, make_class
from typing import Dict, Type, Optional, List, TypeVar, Generic
from cattr import structure

@attrs(auto_attribs=True)
class M(Generic[Q]):
    m: Q

@attrs(auto_attribs=True)
class N(Generic[P]):
    n: P

data2 = {
    "n": {
        "m": "test"
    }
}

a = structure(data2, N[M[str]])
print(a)
```

This allows for deeply nested types and handling generics properly based on the type entered at the structure level. It handles duplicate Type variable names like `T` being used everywhere in the code base, and it supports generics in every sub structure method I have been able to test (unions, list, sets, etc.) to build up an attrs class.

The only issue with this PR is 2 things:
* This is a breaking change, in order for this to work we need to have the mapping information of what the type `T~` actually is suppose to map too, so all structure hooks need to accept 3 variables instead of 2
* All current tests are passing, but I have absolutely zero idea about how to make hypotheses generate random classes and their types to test structuring. I was attempting to look at the nested_classes strategy but I am not familiar enough with hypothesis to get this going.

If some static tests are acceptable, I can easily add those in but I will need some assistance sorting out the hypothesis portion to get tests in line with the current code base.


Tinche:
Hi,

I agree support for this will be useful. All cattrs functionality going forward, though, should be in generated structure/unstructure methods. This functionality is in a branch (https://github.com/Tinche/cattrs/tree/generation) which I plan on merging in very soon, and this functionality will become default in one of the future versions, when it's ironed out enough. So ideally this work would be in the functions there (I guess the structuring function, since that's where the generics are used).

If it's complicated, don't worry about Hypothesis, we can just use a couple static tests and I can fill in the gaps later.

wakemaster39:
Thanks @Tinche. 

I have gone in and rebased this against generation and added support for generics in generation. There is one nasty name conversion string that I am not sure how to best handle as I don't know how unique things need to be but the code is there and working.

src/cattr/gen.py
```diff
@@ -92,9 +95,31 @@ def make_dict_unstructure_fn(cl, converter, omit_if_default=False, **kwargs):
...
+            mapping = converter._generate_mapping(base, mapping)
+            break
+
+    if isinstance(cl, TypeVar):
+        cl = getattr(mapping, cl.__name__, cl)
+
     cl_name = cl.__name__
     fn_name = "structure_" + cl_name
-    globs = {"__c_s": converter.structure, "__cl": cl}
+    for p in getattr(cl, "__parameters__", ()):
```
Tinche, line 115: This block is to come up with a pretty name for the function? Can leave a comment.

src/cattr/gen.py
```diff
@@ -92,9 +95,31 @@ def make_dict_unstructure_fn(cl, converter, omit_if_default=False, **kwargs):
...
+        # This is nasty, I am not sure how best to handle `typing.List[str]` or `TClass[int, int]` as a parameter type here
+        name_base = getattr(mapping, p.__name__)
+        name = getattr(name_base, "__name__", str(name_base))
+        name = re.sub("[\[\.\] ,]", "_", name)
```
Tinche, line 119: This line raises a Flake8 warning on my end.

Tinche (changes requested):
Ok, cool.

What I had in mind was removing basically all changes from `converters.py`, and having the logic only in `gen.py`. If this makes you think you will need to manually generate structure hooks for all your generic classes - one of the next steps is generating these hooks automatically in a converter subclass. But that comes a little later.

src/cattr/gen.py
wakemaster39: this should be ripped out and replaced with something nicer unless this way of generating a name is ok and I can replace it.

wakemaster39:
Right now because this class only does the initial entry point and doesn't handle things further down the mapping functions in converters is still needed or else nested generics explode. 

So to remove the generic changes we will need that convert subclass implemented.

wakemaster39:
I reached into my bag of code that this library inspired and found a way to remove all the generic stuff today and in my opinion don't need subclasses, but I don't 100% follow what you are after. The cost depends on how much you hate recursion.

There are broken tests and shit now, I didn't want to put in the effort until you bought off on the way this works. This is what a working call chain looks like. This change makes `def structure_attrs_fromdict(self, obj, cl, mapping)` obsolete as well.

```python
def test_able_to_structure_deeply_nested_generics_gen(converter):
    cl = TClass[TClass[TClass[int, int], int], int]
    s_fn = make_dict_structure_fn(cl, converter)
    for t, f in s_fn:
        converter.register_structure_hook(t, f)
    result = TClass(TClass(TClass(1, 2), 3), 4)

    res = converter.structure(asdict(result), cl)

    assert res == result
```

Tinche:
@wakemaster39 Note: I haven't had time to go over your commits yet. Broken tests are fine until the feature is done, of course.

So my idea was to have a converter subclass with a different `structure_attrs_fromdict`, which would generate a specialized hook once it saw a class for the first time. The converter machinery would take care of dealing with nested classes. Doing recursion in the hook generator itself wouldn't be necessary. The generated hook just needs to pass the correct type to `converter.structure`. Let me know if this doesn't work and we can brainstorm a different approach.

src/cattr/converters.py
```diff
@@ -240,7 +245,7 @@ def _structure_default(self, obj, cl):
         Bare optionals end here too (optionals with arguments are unions.) We
         treat bare optionals as Any.
         """
-        if cl is Any or cl is Optional:
+        if cl is Any or cl is Optional or cl is None:
```
wakemaster39, line 248: This feels suspect, why is the type none from hypothesis?

wakemaster39:
@Tinche I think this is what you were after. I didn't kill out the structure from tuple but it will need to die as well to get generic support like this. 

`make_dict_structure_fn` at this point really feels like it should be `_make_dict_structure_fn` and have a `make_and_register_dict_structure_fn` since this stuff all injects itself recursively into the converter now.

This certainly cleaned up the entire logic though.

src/cattr/converters.py
```diff
@@ -1,3 +1,4 @@
+from copy import copy
```
Tinche, line 1: Unused import?

src/cattr/converters.py
Tinche: Interesting, you're saying Hypothesis generates Nones for the type annotations? Not sure why off the top of my head. Which tests get it?

Tinche (changes requested):
Good work.

I agree converters and the generating functions should merge down the line. They're interdependent and probably should be one module. But we can postpone this.

I think this generating approach will become the default, but maybe not in the first release/iteration. So I would ask you to basically revert the Converter class in converters.py, and subclass it it gen.py (ugh I'm bad with names, GenConverter?) with these changes. It's ok if the old converter doesn't support generics or postponed annotations - this will get users to try and test the new converter. 

Then we add documentation and release this.

wakemaster39:
All of the structure attrs tests fail when I did this, it was something I fought originally trying to get the unit tests going too. It appears that intergers the type is set to None. I am not sure if there was others, those were just the times I ran into.

I have two alternatives I want run by you @Tinche before I make this change just based on how I have ended up using cattrs.

1. Branch current stable and make this 2.0.0-beta.
2. Make a flag on the converter to flip it between the two strategies

I probably didn't follow the best advice in the world and just think of `structure` and `unstructure` as functions and ignore the fact that they are really a field on a singleton. This leads to me having `from cattr import structure` in a lot of places in my code base.

This makes it pretty impossible to try out a different converter, it is sort of an all or nothing for me. Creating a beta package where this is default would let me pin the beta version and my code should just work using the new strategy.

The alternative of making a flag also solves this problem. As long as I imported `set_use_get()` or something like that and called it once my code start using the strategy and be magic again. 

Thinking about it as I write this, the biggest reason that either of these two things are needed is because the `is_generic` function must be the absolute last thing that gets checked or else it breaks lists and all sorts of other things. So I have a third option.

3. wire `_structure_default` up to use the new generator logic.

The advantage here is all existing code would still run, using the existing logic. If someone passed in a generic it would just work as well. If someone passed in a generic would its args, it would still fail like normal. And if someone wanted to test this as the default mechanism they would just need to reregister the `_is_attrs_class` hook.

I think this one makes the most sense to me. Is none of these work I'll implement your change and keep using a fork for my stuff where this is default until it is available via the stupid import choice that I made.

Tinche:
Couldn't you also easily monkeypatch `cattr.structure` and `cattr.unstructure` with a new singleton?

In any case, I'm OK with option 3.

The reason for a subclass is this subclass should be able to take parameters to customize actually generating the hooks. I'm not yet quite sure what this would look like.

wakemaster39:
I could yes monkey patch it, I am just thinking barrier to entry to get people to try it out and test it. Was just thinking "add `cattr.XXX()`" to your code would be an easier sell than importing cattrs and overriding some things as then order of when that override occurred could matter.

Either way, I think what is implemented here is the best of the choices anyways.

Tinche:
Thanks for your work on this! I will pull this in now, keep working on it a little and cut a release next week.

PR #103, opened 2020-10-30
Set Explicit requirement for Python>=3.7

kaxil:
Check https://packaging.python.org/guides/dropping-older-python-versions/ for more details

The latest cattrs 1.1.0 removed support for Python 3.6 but it since most of the libraries like Airflow (I am the maintainer and PMC Member of the Airflow Project) have an open bound for cattrs in the dependency, new users trying to setup airflow breaks (Issue: https://github.com/apache/airflow/issues/11965).

Setting an explicit `python_requires` means Python 3.6 users won't be able to download the new cattrs version and fallback to the older version that supports Python 3.6 and hence it will work without breaking Airflow too.

Thanks for creating and maintaining this library, appreciate all your effort, hope this PR is useful

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     license="MIT license",
     zip_safe=False,
     keywords="cattrs",
-    python_requires='~=3.7',
+    python_requires="~=3.7",
     classifiers=[
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
     license="MIT license",
     zip_safe=False,
     keywords="cattrs",
+    python_requires='~=3.7',
     classifiers=[
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",


kaxil:
cc @Tinche

jqmp:
I ran into this problem too (we depend on `cattrs` in the [Bionic](https://github.com/square/bionic) library). It seems to me that someone will also need to delete version 1.1.0 from PyPI, right? Otherwise, `pip` will still happily install 1.1.0 on Python 3.6.

Thanks @Tinche for the great library, and @kaxil for submitting this fix!

Tinche:
Thanks, let's get this applied! I will release this as 1.1.1 and delete 1.1.0.

Tinche:
Release 1.1.0 yanked!

kaxil:
Thanks for the quick release @Tinche , I can confirm this works now
```
root@283e9b83e39d:/# python --version
Python 3.6.12

root@283e9b83e39d:/# pip install 'cattrs==1.1.1'
ERROR: Could not find a version that satisfies the requirement cattrs==1.1.1 (from versions: 0.5.0, 0.6.0, 0.7.0, 0.8.0.dev0, 0.8.0, 0.8.1, 0.9.0, 0.9.2, 1.0.0rc0, 1.0.0, 1.1.0)
ERROR: No matching distribution found for cattrs==1.1.1
```

potiuk:
Nice @kaxil!

PR #114, opened 2021-01-20
use lru_cache decorator on dispath

ItayGoren:


Tinche:
Tests are failing. You probably need to remove the slot.

PR #117, opened 2021-01-24
fix MultiStrategyDispatch to work with new GenConverter and attrs inheritance

### LINKED ISSUE #116: GenConverter doesn't support inheritance
* cattrs version: 1.1.1
* Python version: 3.7
* Operating System: Mac

### Description

The old converter supports attrs classes which inherit from each other because it doesn't use MRO-based singledispatch per type to dispatch its attrs classes - therefore, when a subclass B is encountered, it is dynamically structured based on the actual type and the attributes available in the dict.

However, the new GenConverter creates and registers a single-dispatched function as soon as it sees a new type. And functools.singledispatch seems to do some MRO stuff and identifies any subclass as "being" its parent class, therefore skipping the code generation and registration process that cattrs would otherwise do for any new subclass.

### What I Did

```
from cattr.converters import GenConverter
import attr


@attr.s(auto_attribs=True)
class A:
    i: int


@attr.s(auto_attribs=True)
class B(A):
    j: int


def test_inheritance():
    converter = GenConverter()

    # succeeds
    assert A(1) == converter.structure(dict(i=1), A)

    # fails
    assert B(1, 2) == converter.structure(dict(i=1, j=2), B)
```

the failure is 
```
E       AssertionError: assert B(i=1, j=2) == A(i=1)
E        +  where B(i=1, j=2) = B(1, 2)
```

essentially, cattrs/singledispatch chooses A and then runs the A structuring code, rather than generating a new method.

How to fix this? I know singledispatch should be fairly well optimized, so it would be un-fun to drop it entirely. But since this is a breaking change from the old converter to the new converter, I feel like we should try to do something.

I wonder if it might be just as well to use a strict type map looking in the multistrategy dispatch before we hit the singledispatch function. Essentially, if there's an exact match, use that before even trying singledispatch. If there isn't, *and* the type is an attrs class, _skip_ singledispatch so that the class is "caught" by the fallback function dispatch. 

If you're willing to accept this approach, I found a patch that works nicely. I'll probably put up a PR in the next couple of days.

petergaultney:
Because the GenConverter specifically handles generating code for
attrs classes, and because it registers hooks that don't require
function dispatch, singledispatch was preventing subclasses from
having code generated for them, because they would trigger the
previously-generated base class's structure/unstructure code.

This changes MultiStrategyDispatch to allow for a class-based hook to
specifically avoid single-dispatch. Since we know that a given attrs
class can always have code generated for it, we don't really want to
share dispatch across subclasses - like the original Converter, we
want to make sure we're structuring each individual attrs class as its
own separate type.

fixes #116

### DIFF
--- a/src/cattr/multistrategy_dispatch.py
+++ b/src/cattr/multistrategy_dispatch.py
@@ -15,16 +15,24 @@ class _DispatchNotFound(object):
 class MultiStrategyDispatch(object):
     """
     MultiStrategyDispatch uses a
-    combination of FunctionDispatch and singledispatch.
+    combination of exact-match dispatch, singledispatch, and FunctionDispatch.
 
-    singledispatch is attempted first. If nothing is
-    registered for singledispatch, or an exception occurs,
+    Exact match dispatch is attempted first, based on a direct
+    lookup of the exact class type, if the hook was registered to avoid singledispatch.
+    singledispatch is attempted next - it will handle subclasses of base classes using MRO
+    If nothing is registered for singledispatch, or an exception occurs,
     the FunctionDispatch instance is then used.
     """
 
-    __slots__ = ("_function_dispatch", "_single_dispatch", "dispatch")
+    __slots__ = (
+        "_direct_dispatch",
+        "_function_dispatch",
+        "_single_dispatch",
+        "dispatch",
+    )
 
     def __init__(self, fallback_func):
+        self._direct_dispatch = dict()
         self._function_dispatch = FunctionDispatch()
         self._function_dispatch.register(lambda _: True, fallback_func)
         self._single_dispatch = singledispatch(_DispatchNotFound)
@@ -32,6 +40,9 @@ class MultiStrategyDispatch(object):
 
     def _dispatch(self, cl):
         try:
+            direct_dispatch = self._direct_dispatch.get(cl)
+            if direct_dispatch:
+                return direct_dispatch
             dispatch = self._single_dispatch.dispatch(cl)
             if dispatch is not _DispatchNotFound:
                 return dispatch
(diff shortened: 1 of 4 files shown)


src/cattr/multistrategy_dispatch.py
```diff
@@ -15,34 +15,50 @@ class _DispatchNotFound(object):
...
-    __slots__ = ("_function_dispatch", "_single_dispatch", "dispatch")
+    __slots__ = (
+        "_direct_dispatch",
+        "_function_dispatch",
+        "_single_dispatch",
+        "dispatch",
+    )
 
     def __init__(self, fallback_func):
+        self._direct_dispatch = dict()
```
Tinche, line 35:
```suggestion
        self._direct_dispatch = {}
```
For consistency

src/cattr/multistrategy_dispatch.py
```diff
@@ -15,34 +15,50 @@ class _DispatchNotFound(object):
...
         self._function_dispatch = FunctionDispatch()
         self._function_dispatch.register(lambda _: True, fallback_func)
         self._single_dispatch = singledispatch(_DispatchNotFound)
         self.dispatch = lru_cache(maxsize=None)(self._dispatch)
 
     def _dispatch(self, cl):
         try:
+            direct_dispatch = self._direct_dispatch.get(cl)
+            if direct_dispatch:
```
Tinche, line 44:
```suggestion
            if direct_dispatch is not None:
```

Tinche (changes requested):
Please add a HISTORY entry with a link to the PR and issue and I'll merge this in :)

src/cattr/multistrategy_dispatch.py
Tinche, line 44:
```suggestion
            if direct_dispatch is not None:
```

petergaultney:
i put the entry in 1.2.0 but if you decide to do a bugfix release first, let me know and I can change it to a 1.1.3.

Tinche (approved):
Thanks for this. I might add a few tests and then release this as 1.2.0 in a few days.

Tinche:
I had to tweak this a little so that direct dispatch is consulted after the singledispatch, not before, to get some tests to pass.

PR #134, opened 2021-03-16
Add smoke tests for nested attrs-dataclasses

asford:
https://github.com/Tinche/cattrs/issues/43

Minor fixes to converters.py for some basic conversion smoke tests.


Tinche:
Hm I figured I would limit this functionality to the `GenConverter`, but I guess it makes sense to retrofit the legacy Converter too since that's currently used for the global converter.

Tinche:
Which reminds me, can you parametrize these smoke tests to test both converters?

asford:
👍 That'd be ideal for our case, as we've folks who just use the "out of box" settings for a lot of applications.

asford:
Added parameterization, obviously feel free to slice and dice the PR as makes sense to you.

Tinche:
Thanks!

PR #139, opened 2021-03-26
Structure using the value of attr.ib converter, if defined

natemcmaster:
Resolves https://github.com/Tinche/cattrs/issues/138

If an attrs class has defined a converter with  `attr.ib`, fallback that instead if dispatching to cattrs converters fails to find a handler.

Also, add a new option `prefer_attrib_converters` to use them if defined instead of handlers registered in cattrs.

### DIFF
--- a/src/cattr/gen.py
+++ b/src/cattr/gen.py
@@ -1,3 +1,4 @@
+import functools
 import re
 from dataclasses import is_dataclass
 from typing import Any, Optional, Type, TypeVar
@@ -6,6 +7,7 @@ import attr
 from attr import NOTHING, resolve_types
 
 from ._compat import adapted_fields, get_args, get_origin, is_generic
+from .errors import StructureHandlerNotFoundError
 
 
 @attr.s(slots=True, frozen=True)
@@ -162,10 +164,17 @@ def make_dict_structure_fn(
         # For each attribute, we try resolving the type here and now.
         # If a type is manually overwritten, this function should be
         # regenerated.
-        if type is not None:
+        if converter._prefer_attrib_converters and a.converter is not None:
+            # The attribute has defined its own conversion, so pass
+            # the original value through without invoking cattr hooks
+            handler = _passthru
+        elif type is not None:
             handler = converter._structure_func.dispatch(type)
         else:
             handler = converter.structure
+
+        if not converter._prefer_attrib_converters and a.converter is not None:
+            handler = _fallback_to_passthru(handler)
 
         struct_handler_name = f"__cattr_struct_handler_{an}"
         globs[struct_handler_name] = handler
@@ -199,6 +208,21 @@ def make_dict_structure_fn(
     eval(compile("\n".join(total_lines), "", "exec"), globs)
 
     return globs[fn_name]
+
+
+def _passthru(obj, _):
+    return obj
+
+
+def _fallback_to_passthru(func):
+    @functools.wraps(func)
+    def invoke(obj, type_):
+        try:
+            return func(obj, type_)
+        except StructureHandlerNotFoundError:
+            return obj
+
+    return invoke
 
 
 def make_iterable_unstructure_fn(cl: Any, converter, unstructure_to=None):
--- a/src/cattr/gen.py
+++ b/src/cattr/gen.py
@@ -135,6 +135,7 @@ def make_dict_structure_fn(
     converter,
     _cattrs_forbid_extra_keys: bool = False,
     _cattrs_use_linecache: bool = True,
+    _cattrs_prefer_attrib_converters: bool = False,
     **kwargs,
 ):
     """Generate a specialized dict structuring function for an attrs class."""
@@ -187,7 +188,7 @@ def make_dict_structure_fn(
         # For each attribute, we try resolving the type here and now.
         # If a type is manually overwritten, this function should be
         # regenerated.
-        if converter._prefer_attrib_converters and a.converter is not None:
+        if _cattrs_prefer_attrib_converters and a.converter is not None:
             # The attribute has defined its own conversion, so pass
             # the original value through without invoking cattr hooks
             handler = _passthru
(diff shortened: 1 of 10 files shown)


Tinche:
Cool, thanks for working on this.

Don't forget to add a changelog entry, and keep it updated.

This will need to be a flag though, and it needs to be disabled by default due to a) backwards compatibility, and b) it breaks a very common use case:

```
from typing import Sequence

import attr

from cattr import GenConverter

c = GenConverter()


@attr.define
class Inner:
    a: int


@attr.define
class Outer:
    inners: Sequence[Inner] = attr.field(converter=tuple)


print(c.structure({"inners": [{"a": 1}]}, Outer))
```
So this flag will need documentation too.

I wonder if a good heuristic would be: while generating the structure function for a class, if a field does not have a handler registered, but it does have a converter, use the converter instead. That might keep backwards compatibility somewhat and help with your use case (and the example would still work).

natemcmaster:
So, before I implement anything else, here are some alternatives

**Option 1** - behavior as I've implemented now, but add a flag which could be used to opt-out
```py
# example of opting out
c = Converter(use_attrib_converters=False)
c.structure(...)
```

**Option 2** - make this behavior an opt-in
```py
# example of opting in
c = Converter(use_attrib_converters=True)
c.structure(...)
```

**Option 3** - no new API, fallback to attr.ib converters if no handler is registered


**Option 4** - fallback to attr.ib converters if no handler is found, but with an option to give attr.ib converters higher priority than registered handlers
```py
c = Converter(prefer_attrib_converters=True)
c.structure(...)
```

Benefits of option 1 and 4 are that you could register an attribute converter which then calls cattr.structure.

-----

I personally prefer option 1. As a user, this fits with what I was expecting to happen when registering an attr.ib converter.

I see the sense in the backwards compatible issue. What are your thoughts doing one of the other options in a 1.x update, and making using attr.ib converters the default in a new major version later?

Tinche:
My order of preference would be 3, 4 and then 2. Option 1 by default breaks a use case that I'd like to support (referenced in my first comment).

natemcmaster:
Thanks for the feedback, @Tinche. I've implemented option 3 from above and updated documentation. Let me know what you think!

Tinche:
@natemcmaster Thanks! I have a couple busy days ahead so I can't look at this right away, but will as soon as I can!

src/cattr/gen.py
```diff
@@ -162,11 +164,18 @@ def make_dict_structure_fn(
         # For each attribute, we try resolving the type here and now.
         # If a type is manually overwritten, this function should be
         # regenerated.
-        if type is not None:
+        if converter._prefer_attrib_converters and a.converter is not None:
```
Tinche, line 167: Instead of reading this from the converter, let's make it an argument (`_cattrs_prefer_attrib_converters`), and pass it in when calling this from the `GenConverter`.

src/cattr/gen.py
```diff
@@ -162,11 +164,18 @@ def make_dict_structure_fn(
...
+            # The attribute has defined its own conversion, so pass
+            # the original value through without invoking cattr hooks
+            handler = _passthru
```
Tinche, line 170: I try generating very optimized code here. I would set the handler to `None` and below, when generating the structuring code, check if it's `None` and not even call anything, just grab the key from the dictionary and put it into the `res` dict directly.

src/cattr/gen.py
```diff
@@ -162,11 +164,18 @@ def make_dict_structure_fn(
...
+        elif type is not None:
             handler = converter._structure_func.dispatch(type)
         else:
             handler = converter.structure
 
+        if not converter._prefer_attrib_converters and a.converter is not None:
+            handler = _fallback_to_passthru(handler)
```
Tinche, line 177: I don't think we need to use `_fallback_to_passthru`, we can know here and now if we will need to passthrough, right? If the handler is equal to `converter._structure_default`, the type is not `is_generic` and calling `_structure_default` raises an exception. It's a little dirty but should work.

Ideally the `structure_default` handler can be refactored to only contain the error logic, but that might be a different task.

Tinche (changes requested):
Left some comments, also don't forget to modify the changelog. Thanks!

src/cattr/gen.py
natemcmaster: Sorry, I don't quite follow what you're proposing the code should do differently here.

natemcmaster:
Updated per your feedback, except for the one piece I didn't understand.

Side note -- some thoughts about contributing to this project -- I'm not quite sure if I've done the right things, which will likely lengthen the feedback loop here. For example, the contributing.rst guide seems out of date with the changes to add poetry, use GitHub actions, and doesn't mention if I need to run formatters/linters (seems the CI may not check formatting.)

Anyways, lmk how this next draft works. Also, if you have a really specific idea of how something should work, feel free to edit my PR and merge without waiting for me.

Tinche:
> Side note -- some thoughts about contributing to this project -- I'm not quite sure if I've done the right things, which will likely lengthen the feedback loop here. For example, the contributing.rst guide seems out of date with the changes to add poetry, use GitHub actions, and doesn't mention if I need to run formatters/linters (seems the CI may not check formatting.)

You're right. I would be happy to merge a PR improving this :)

Tinche:
I think this looks good, there's a small refactor I plan on making after this is merged to make it a little cleaner, but you shouldn't worry about it. Fix the conflicts and I shall merge.

Thanks a lot!

natemcmaster:
Thanks @Tinche. I believe there are no more merge conflicts with the master branch. Thanks for the additional review :)

Tinche:
Screenshot 2021-06-18 at 17 24 12

Still can't merge for some reason

natemcmaster:
@Tinche does squash merge work instead?

Tinche:
Looks like it worked! Thanks.

PR #142, opened 2021-04-02
Forbid extra fields when structuring dictionaries into attrs classes

### LINKED ISSUE #101: [Feature] Throw error when structuring dict w/ extra keys
* cattrs version: 1.0.0
* Python version: Python 3.6
* Operating System: Linux (Fedora)

### Description

Right now, `cattr.structure` will silently ignore *extra* keys when structuring into an attrs object:

```python
In [1]: import attr
   ...: import cattr
   ...: 
   ...: @attr.s
   ...: class A:
   ...:     a: int = attr.ib()
   ...: 
   ...: cattr.structure({"a": 1, "b": 2}, A)
Out[1]: A(a=1)
```

I assume this is intended behavior (since I believe this changed since cattrs 0.9), but it'd be nice to have an option to throw an error on extra keys instead.

In our particular use-case, we are using cattrs and attrs to verify some configuration files we are using, some of which have default keys:

```
@attr.s
class Config:
    some_key: int = attr.ib(default=0)
```

b/c of this "silently allow extra keys" behavior, `cattr.structure` doesn't catch typos (e.g. `some_key` vs `sme_key`), because it will just ignore the incorrectly-typed key and use the default value for the substitute.

bkurtz:
Would resolve #101.  I'm assuming this will need some further work (e.g. new tests; fixing other tests this breaks, etc), but wanted to go ahead and get it up for a first review to see if the approach is reasonable.

### DIFF
--- a/src/cattr/gen.py
+++ b/src/cattr/gen.py
@@ -109,7 +109,9 @@ def generate_mapping(cl: Type, old_mapping):
     return cls(**mapping)
 
 
-def make_dict_structure_fn(cl: Type, converter, **kwargs):
+def make_dict_structure_fn(
+    cl: Type, converter, _cattrs_forbid_extra_keys: bool = False, **kwargs
+):
     """Generate a specialized dict structuring function for an attrs class."""
 
     mapping = None
@@ -181,6 +183,16 @@ def make_dict_structure_fn(cl: Type, converter, **kwargs):
                 f"    res['{ian}'] = {struct_handler_name}(o['{kn}'], __c_t_{an})"
             )
     lines.append("    }")
+    if _cattrs_forbid_extra_keys:
+        allowed_fields = {a.name for a in attrs}
+        globs["__c_a"] = allowed_fields
+        post_lines += [
+            "  unknown_fields = set(o.keys()) - __c_a",
+            "  if unknown_fields:",
+            "    raise Exception(",
+            f"      'Extra fields in constructor for {cl_name}: ' + ', '.join(unknown_fields)"
+            "    )",
+        ]
 
     total_lines = lines + post_lines + ["  return __cl(**res)"]
 
--- a/tests/metadata/test_genconverter.py
+++ b/tests/metadata/test_genconverter.py
@@ -18,7 +18,7 @@ from hypothesis.strategies import booleans, sampled_from, lists
 
 from cattr import GenConverter as Converter
 from cattr import UnstructureStrategy
-from cattr.gen import override
+from cattr.gen import make_dict_structure_fn, override
 
 from . import (
     nested_typed_classes,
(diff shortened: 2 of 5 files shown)


src/cattr/gen.py
```diff
@@ -181,6 +183,17 @@ def make_dict_structure_fn(cl: Type, converter, **kwargs):
                 f"    res['{ian}'] = {struct_handler_name}(o['{kn}'], __c_t_{an})"
             )
     lines.append("    }")
+    if forbid_extra_fields:
+        post_lines.append("  allowed_fields = {")
```
bkurtz, line 187: Wasn't sure how much of a (non-)goal it is to have tidy generated code.  For example, I could probably have written
```python
post_lines.append(f"  allowed_fields = {{a.name} for a in attrs:r}")
```
which is shorter and maybe a little cleaner from the point of view of this function, but would likely lead to a very long line in the generated code in many cases.

src/cattr/gen.py
```diff
@@ -109,7 +109,9 @@ def generate_mapping(cl: Type, old_mapping):
     return cls(**mapping)
 
 
-def make_dict_structure_fn(cl: Type, converter, **kwargs):
+def make_dict_structure_fn(
+    cl: Type, converter, forbid_extra_fields: bool = False, **kwargs
```
Tinche, line 113: The argument at this layer should be a little obfuscated, since it might shadow an attribute name you're trying to override (in **kwargs). I'd change it to `_cattrs_forbid_extra_fields`. Note that we can add this flag on the `GenConverter` level with a better name.

src/cattr/gen.py
```diff
@@ -181,6 +183,17 @@ def make_dict_structure_fn(cl: Type, converter, **kwargs):
                 f"    res['{ian}'] = {struct_handler_name}(o['{kn}'], __c_t_{an})"
             )
     lines.append("    }")
+    if forbid_extra_fields:
+        post_lines.append("  allowed_fields = {")
+        post_lines += [f"    '{a.name}'," for a in attrs]
+        post_lines += [
+            "  }",
+            "  unknown_fields = set(o.keys()) - allowed_fields",
+            "  if unknown_fields:",
```
Tinche, line 192: I would condense this into a single line
```suggestion
            "  if o.keys() - allowed_fields:",
```

src/cattr/gen.py
```diff
@@ -181,6 +183,17 @@ def make_dict_structure_fn(cl: Type, converter, **kwargs):
...
+            "    raise KeyError(",
```
Tinche, line 193: I wouldn't use a `KeyError` here. First of all, `KeyError` is raised when a key is missing from a dictionary, not when it's extra :)

Second, `KeyError` is a built-in Python exception. User libraries are supposed either throw `Exception` or something inheriting from `Exception` (see https://docs.python.org/3/library/exceptions.html#Exception). Ideally we would have a `StructureError` and you could inherit from that, but we don't currently and that should be a different task, so just throw `Exception` for now.

Tinche (changes requested):
Hello,

thanks for your work on this, let's see about tidying this up and getting it into the next release.

I would also possibly add a parameter for the `GenConverter`, similar to `omit_if_default`, that would be used for generating all classes here.

src/cattr/gen.py
bkurtz: Ah, I wondered about that.  Will do.

src/cattr/gen.py
bkurtz: Except then I can't re-use it to list the unknown fields in the exception message.  Or would you prefer to recalculate it again?

src/cattr/gen.py
bkurtz: Hm, I think `KeyError` counts as "something inheriting from `Exception`", but agreed that custom exception subclasses would be best, and that a `KeyError` is a little weird.
```python
try:
    raise KeyError('key error')
except Exception:
    print("caught your key error")
```
I was thinking of it as the key being missing from the *class*, rather than from the dictionary.  But I'm happy to put in `Exception` for now if that's your preference!

bkurtz:
While working on this round of updates, I noticed that you'd suggested `forbid_extra_keys` as the `GenConverter` option and `_cattr_forbid_extra_fields` as the argument to `make_dict_structure_fn`.  I figured that they should probably match (fields vs keys), and I felt sorta like "keys" was better, so I went with that, but happy to revert it if you prefer the other way.

Started thinking about tests a little tonight, but would probably appreciate some feedback on what sort of tests are desired here:
- I assume there should be at least one test of the updated `make_dict_structure_fn` directly; should it be
  * a one-off test to show that an error is raised in a single case where there's an extra field (fast)
  * property testing with hypothesis to show that this works on a large variety of cases, e.g. add a character to a field name, double the field name, etc.  (slower, probably harder to write a good test)

  In either case, I'm guessing this test would live in `test_dict_generation.py`?
- I assume there should be at least one test of using the setting with a `GenConverter` object as well, but it wasn't as clear to me where this one should live.

bkurtz:
PS. Thanks for the quick turnaround with a review last night!

src/cattr/gen.py
Tinche: Er good point!

src/cattr/gen.py
Tinche: Yeah, let's switch to a bare `Exception` for now. `KeyError` is a bit of a stretch I think.

Tinche:
> While working on this round of updates, I noticed that you'd suggested `forbid_extra_keys` as the `GenConverter` option and `_cattr_forbid_extra_fields` as the argument to `make_dict_structure_fn`. I figured that they should probably match (fields vs keys), and I felt sorta like "keys" was better, so I went with that, but happy to revert it if you prefer the other way.

Er yeah, keys is better.

> Started thinking about tests a little tonight, but would probably appreciate some feedback on what sort of tests are desired here:
> 
>     * I assume there should be at least one test of the updated `make_dict_structure_fn` directly; should it be
>       
>       * a one-off test to show that an error is raised in a single case where there's an extra field (fast)
>       * property testing with hypothesis to show that this works on a large variety of cases, e.g. add a character to a field name, double the field name, etc.  (slower, probably harder to write a good test)
>       
>       In either case, I'm guessing this test would live in `test_dict_generation.py`?

If you don't have Hypothesis experience it's ok to skip. Yeah, I'd just add a test or two to `test_dict_generation`. Perfectly fine for them to be simple.

>     * I assume there should be at least one test of using the setting with a `GenConverter` object as well, but it wasn't as clear to me where this one should live.

That would be great, yeah. You can add a test to `metadata/test_genconverter.py`.

Thanks!

bkurtz:
Added some tests.  Ended up feeling lazy tonight about tests for `make_dict_structure_fn` since `GenConverter` exercises that anyway, but can come back and add those another night if you think it'd be good to have both.

Started on docs, but not done there.

Tinche:
Cool, no rush. Thanks!

bkurtz:
Okay, I think I'd be ready to call this done.  Let me know if there are other changes you'd like to see.

bkurtz:
Also: I'm accustomed to a merge/squash workflow to keep git history tidy, but have been assuming it would be appropriate to keep the history as-is to make reviewing easy.  LMK if/when it's appropriate to squash.

Tinche (changes requested):
Great job! Left a comment inline.

Also, to save me from opening up a PR to your PR, can you just apply this change:

```
 if _cattrs_forbid_extra_keys:
        allowed_fields = {a.name for a in attrs}
        globs["__c_a"] = allowed_fields
        post_lines += [
            "  unknown_fields = set(o.keys()) - __c_a",
            "  if unknown_fields:",
            "    raise Exception(",
            f"      'Extra fields in constructor for {cl_name}: ' + ', '.join(unknown_fields)"
            "    )",
        ]
```
(i.e. precalculate `allowed_fields` and make it available in the globals).

tests/metadata/test_genconverter.py
```diff
@@ -59,6 +59,96 @@ def test_simple_roundtrip_defaults(attr_and_vals, strat):
...
+    converter = Converter(unstruct_strat=strat, forbid_extra_keys=True)
+    cl, vals = cls_and_vals
+    inst = cl(*vals)
+    unstructured = converter.unstructure(inst)
+    assert "Hyp" not in repr(unstructured)
+    assert inst == converter.structure(unstructured, cl)
+
+
+@given(simple_typed_attrs(defaults=True), unstructure_strats)
+def test_simple_roundtrip_defaults_with_extra_keys_forbidden(
```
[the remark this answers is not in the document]
bkurtz: I think what I had in mind at the time was to make sure that it hadn't somehow made it error in the wrong order (i.e. when there are keys present in the structured object that aren't present in the dict), but given the current implementation I think that's not really a concern, and `test_forbid_extra_keys_nested_override` would probably catch that in a simple case anyway, so I've gone ahead and deleted.  Definitely nice to avoid having too many tests with hypothesis - can really increase the testing time.

bkurtz:
Should be set I think.  Pre-calculating the allowed fields is a nice touch - thanks for that pointer.
Let me know if I should squash changes or anything like that.

Tinche:
I don't really have a policy for squashing, feel free if you feel like it. I will take another look at this in the next day or two and maybe merge it it. Note that GitHub's telling me this branch cannot be rebased to due conflicts for some reason

bkurtz:
Squashed.

Tinche:
Great works, thanks a lot!

PR #145, opened 2021-05-04
Improve type inference in register_{un,}structure_hook

bluetech:
Previously, the following line

```py
converter.register_unstructure_hook(datetime.date, lambda dt: dt.isoformat())
```

causes a mypy error

```
Cannot infer type argument 1 of "register_unstructure_hook" of "Converter"  [misc]
```

because the type of `register_unstructure_hook` was

```py
def register_unstructure_hook(self, cls: Any, func: Callable[[T], Any]) -> None: ...
```

and the `lambda` does not specify the `T`.

A workaround is to use a `def` instead of the `lambda` which specifies
the argument type, thus allowing mypy to infer `T`, but this is a bit
annoying.

Instead, change the `cls: Any` to `cls: Type[T]`, allowing mypy to bind
`T` to the given `cls` already. This also verifies that `cls` and
`func` are consistent.

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -162,7 +162,7 @@ class Converter(object):
         )
 
     def register_unstructure_hook(
-        self, cls: Any, func: Callable[[T], Any]
+        self, cls: Type[T], func: Callable[[T], Any]
     ) -> None:
         """Register a class-to-primitive converter function for a class.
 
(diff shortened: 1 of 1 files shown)


bluetech:
I'm not sure if the `Any` was intentional or not, but easy enough to send a PR to find out :)

If this is acceptable, one way to "test" this is to add an empty file which isn't executed by pytest but is checked by mypy, and ensure that it's free of mypy errors. I can add such a file if you'd like.

Tinche:
I think I've relaxed the type signatures since `Type[C]` doesn't work for a bunch of stuff (like tuples, optionals, unions).

bluetech:
Yep, makes sense. I tried some overload tricks but didn't work. Well, no lambda for me I guess :)

Tinche:
There was actually some work being done on typing-sig to enable this, but I believe it paused at the moment. https://mail.python.org/archives/list/typing-sig@python.org/thread/7TDCBWT4RAYDJUQJ3B5NKXTQDUO5SIW2/ I believe?

PR #146, opened 2021-05-05
Mark as requiring Python 3.7 and higher

ikonst:
The incompatible change seems to have been introduced in 1.1.0 (in https://github.com/Tinche/cattrs/pull/98) and we cannot fix those packages retroactively but we can at least mark it going forward.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ known_first_party = ["cattr"]
 [tool.poetry]
 name = "cattrs"
 version = "1.7.0dev0"
+requires-python = ">=3.7"
 description = "Composable complex class support for attrs and dataclasses."
 authors = ["Tin Tvrtkovic <tinchester@gmail.com>"]
 license = "MIT"


Tinche:
Ooph sorry, thanks!

Tinche:
Hm this breaks the packaging, I believe this info was already present here: https://github.com/Tinche/cattrs/blob/8a3d6b5464fad202b1d214b33001dce8a7a40a5a/pyproject.toml#L27

Tinche:
Screenshot 2021-05-05 at 12 56 33

Looks fine to me on PyPI.

Tinche:
Yeah, in fact 1.1.0 was yanked from PyPI (for this reason I believe). https://pypi.org/project/cattrs/#history

ikonst:
Ahh, nice. I've missed the fact it's yanked because of our internal pypicloud proxy :(

ikonst:
No, wait, ignore that. My pip 19 was too old. pip 21 skipped yanked versions.

PR #152, opened 2021-05-31
Add project URLs for changelog, issues and code

JarnoRFB:
This should make checking for changes in new versions more easy when coming form PyPI. The names where chosen as in attrs (https://github.com/python-attrs/attrs/blob/main/setup.py#L17), except for `repository` which is a poetry built-in.


Tinche:
Nice, thanks!

PR #157, opened 2021-07-10
Various test suite compatibility improvements

mgorny:
1. Install `poetry` inside tox not to require external install of poetry.
2. Support passing additional args to pytest via tox.
3. <del>Skip respective `test_preconf` tests when deps are missing instead of failing.</del> <ins>Move test-related imports to test functions.</ins>
4. <del>Skip bson tests if incompatible `bson` module from pymongo is installed.</del>

### DIFF
--- a/tests/test_preconf.py
+++ b/tests/test_preconf.py
@@ -5,8 +5,6 @@ from json import loads as json_loads
 from typing import Dict, List
 
 from attr import define
-from bson import dumps as bson_dumps
-from bson import loads as bson_loads
 from hypothesis import given
 from hypothesis.strategies import (
     binary,
@@ -22,15 +20,6 @@ from hypothesis.strategies import (
     sets,
     text,
 )
-from msgpack import dumps as msgpack_dumps
-from msgpack import loads as msgpack_loads
-from orjson import dumps as orjson_dumps
-from orjson import loads as orjson_loads
-from tomlkit import dumps as tomlkit_dumps
-from tomlkit import loads as tomlkit_loads
-from ujson import dumps as ujson_dumps
-from ujson import loads as ujson_loads
-from yaml import safe_dump, safe_load
 
 from cattr._compat import (
     Counter,
@@ -194,6 +183,8 @@ def test_stdlib_json(everything: Everything):
     )
 )
 def test_ujson(everything: Everything):
+    from ujson import dumps as ujson_dumps
+    from ujson import loads as ujson_loads
     converter = ujson_make_converter()
     raw = ujson_dumps(converter.unstructure(everything))
     assert (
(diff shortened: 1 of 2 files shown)


Tinche:
Hello, thanks for contributing!

I'll take a modified PR with 1) and 2) from your list since I feel those are improvements. 3) and 4) I feel simply complicate the tests for no reason (the preconf libraries are listed in the dev dependencies, so when would you run the tests without them?).

mgorny:
We're running tests on production systems to ensure that the package will work correctly on them. However, we aren't interested in installing the plethora of dependencies we're never going to use. What's worse is that pymongo includes a 'bson' package that collides with bson, and therefore we can't install bson without breaking pymongo.

Tinche:
> We're running tests on production systems to ensure that the package will work correctly on them. However, we aren't interested in installing the plethora of dependencies we're never going to use. What's worse is that pymongo includes a 'bson' package that collides with bson, and therefore we can't install bson without breaking pymongo.

The cattrs test suite is written to be run using the cattrs dev environment. We aren't interested in supporting other, arbitrary environments :) That said, you could use the newly added additional args to tox to avoid running the preconf tests.

It's unfortunate to hear about pymongo, especially since we use it (via motor) at $DAYJOB. Since the import is the same, there's no way to test both at the same time, right?

mgorny:
> > We're running tests on production systems to ensure that the package will work correctly on them. However, we aren't interested in installing the plethora of dependencies we're never going to use. What's worse is that pymongo includes a 'bson' package that collides with bson, and therefore we can't install bson without breaking pymongo.
> 
> The cattrs test suite is written to be run using the cattrs dev environment. We aren't interested in supporting other, arbitrary environments :) That said, you could use the newly added additional args to tox to avoid running the preconf tests.

Unfortunately, this means we'll have to skip the whole file since all the imports are in the global scope. Could you at least move them to individual test functions to provide the ability to skip these that are irrelevant while running the rest?

> It's unfortunate to hear about pymongo, especially since we use it (via motor) at $DAYJOB. Since the import is the same, there's no way to test both at the same time, right?

Nothing I can think of. This seems to be a big problem, and apparently upstream knows about it:

https://pymongo.readthedocs.io/en/stable/installation.html
https://jira.mongodb.org/browse/PYTHON-757

Unfortunately, they seem to have chosen to claim that the other package was 'unmaintained' (maybe it was at the time? It was in 2014 apparently) and choose to disallow it. I've filed https://jira.mongodb.org/projects/PYTHON/issues/PYTHON-2805 to make them reconsider but I honestly doubt they'll be doing it after claiming the name for so many years.

Tinche:
@mgorny I'm fine with moving imports into individual tests as a compromise!

As for the bson thing, I guess I'll switch the dev dependency to pymongo instead since it has like two orders of magnitude downloads.

mgorny:
Do you want me to update the PR?

Tinche:
For imports inside tests? Yes please

mgorny:
> For imports inside tests? Yes please

Done.

Tinche:
Thanks!

mgorny:
Thanks!

Issue #160, opened 2021-07-20
Need to use a custom namespace when resolving types coming from annotations which were imported in TYPE_CHECKING blocks


dam5h:
Thank you for this awesome library!  It is a treat to use once I wrapped my head around a few things!

* cattrs version: 1.7.1
* Python version: 3.9.6
* Operating System: Linux

### Description

When trying to use annotations from TYPE_CHECKING blocks on `attrs` classes.  I get a error in the `resolve_types` call since the namespace from `__origin__` does not contain those classes.  Since `resolve_types` allows optional custom namespaces to be passed in, I was able to work around the issue by registering a custom namespace and passing that on to `resolve_types`

### What I Did

Forked the repo and added a `register_namespace` function.  I also made a simple test to recreate the issue and pushed both up to my fork.  I will create a draft pull request for illustrative and conversational purposes. 

```
Here is the traceback for the test I created:
../../.pyenv/versions/3.9.6/envs/edgar-3.9/lib/python3.9/site-packages/attr/_funcs.py:386: in resolve_types
    hints = typing.get_type_hints(cls, globalns=globalns, localns=localns)
        attribs    = None
        cls        = <class 'tests.nested_type_check_test.nested_o.Outer'>
        globalns   = None
        localns    = None
        typing     = <module 'typing' from '/home/dave/.pyenv/versions/3.9.6/lib/python3.9/typing.py'>
../../.pyenv/versions/3.9.6/lib/python3.9/typing.py:1424: in get_type_hints
    value = _eval_type(value, base_globals, localns)
        ann        = {'a': 'InnerE', 'b': 'InnerE', 'c': 'InnerE', 'd': 'InnerE'}
        base       = <class 'tests.nested_type_check_test.nested_o.Outer'>
        base_globals = {'Outer': <class 'tests.nested_type_check_test.nested_o.Outer'>, 'TYPE_CHECKING': False, '__builtins__': {'ArithmeticE...'>, ...}, '__cached__': '/home/dave/code/cattrs/tests/nested_type_check_test/__pycache__/nested_o.cpython-39.pyc', ...}
        globalns   = None
        hints      = {}
        include_extras = False
        localns    = None
        name       = 'a'
        obj        = <class 'tests.nested_type_check_test.nested_o.Outer'>
        value      = ForwardRef('InnerE')
../../.pyenv/versions/3.9.6/lib/python3.9/typing.py:290: in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
[...]

PR #161, opened 2021-07-20
DRAFT for illustration: add optional namespace kws to pass into resolve_types

dam5h:
This is related to issue #160 

In the case of type annotations which are imported in a TYPE_CHECKING block
the origin class will not contain them in their runtime namespace so cattrs was
unable to resolve them which structuring and unstructuring.  Added a method to
register a custom namespace with the Converter class which could be used when
resolving classes not in the default namespace.

A simple test of nesting when those annotations of nested types are coming from
type checking blocks and therefore not present at runtime.  The test file is
test_deep_nest_custom_namespace.py

Thank you for this awesome library, I've found it very helpful!

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -76,6 +76,7 @@ class Converter(object):
         "_union_struct_registry",
         "_structure_func",
         "_prefer_attrib_converters",
+        "_globalns",
     )
 
     def __init__(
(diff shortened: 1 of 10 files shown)


dam5h:
See issue #160, this can be resolved by client code without needing to change `cattrs`

Issue #149, opened 2021-05-26
cattr.structure doesn't support generic parameters.


asford:
*Partially a note to myself, if you're open to a PR fixing this.*

`cattr` has solid support for generics in unstructure/structure operations, however it doesn't fully support nested generic types.

Consider instances where a generic attrs class contains a generic collection as a member...

```python
import attr
from typing import Generic, TypeVar

T = TypeVar("T")

@attr.define
class Foo(Generic[T]):
   bar: Dict[str, T]

cattr.structure({"bar" : {"bat": 1}}, Foo[int])
```

...this fails, though one _might_ expect cattr to propagate the `T` typevar into the `Dict[str, T]` generic member for an inferred `Dict[str, int]` field type.

This can *probably* be fixed by mildly extending the existing gen converter logic.

All the `TypeVar`s in the current context are already resolved into `mapping` at:
https://github.com/Tinche/cattrs/blob/71a7ed2b4c72558af77487ae991f19cc5401c378/src/cattr/gen.py#L117-L126

which is then used to map `TypeVar`s to their concrete type in the generated converter at:
https://github.com/Tinche/cattrs/blob/71a7ed2b4c72558af77487ae991f19cc5401c378/src/cattr/gen.py#L155-L161

Here, generic types could be detected via `typing.get_args` or the equivalent `_compat` function.
If the attribute type contains any `TypeVar` args, these could be replaced via `mapping` lookups and an updated type created with `type.copy_with`:

```python
for a in attrs:
    an = a.name
    override = kwargs.pop(an, _neutral)
    type = a.type
    type_args = typing.get_args(type)
    
    if isinstance(type, TypeVar):
        type = getattr(mapping, type.__name__, type)
    elif type_args is not None:
        type_args = [
            getattr(mapping, arg.__name__, arg)
            for ta in type_args
        ]
        type = type.copy_with(
            tuple(
                getattr(mapping, arg.__name__, arg)
                for ta in type_args
            )
        )
```

This will probably be fragile to updates in the `typing` module, so should probably be moved into _compat.

Tinche:
Yeah I'd definitely be open to merging in something that fixes this.

petergaultney:
this may be our last remaining blocker to using the GenConverter - for some reason it seems this was not an issue with the old converter? At least, using the old converter our code doesn't blow up and structures properly, but switching over, things now seem to be broken.

[...]

PR #171, opened 2021-09-27
fix: deeply nested generics

wakemaster39:
provided default structure for none types

#149


petergaultney:
@jwsloan we should try this branch out.

Tinche:
Ah crap, I merged in and deleted the `generic-structuring-fix` branch which closed this. I did some more work on that branch to make the tests pass, see if master works for your use case now, and if not please reopen this against master and I will review it.

Issue #170, opened 2021-09-26
structure/unstructure doesn't support generic nested attrs objects


Savlik:
* cattrs version: 1.8.0
* Python version: 3.8.10
* Operating System: Linux

### Description

I have two nested attrs objects, the outer one has the attribute as generic. I would expect it would unstructure both as if the attribute was specified directly. Am I doing something wrong, or is it a bug? Is there a way to accomplish this? Thanks

### What I Did

```python
from typing import TypeVar, Generic
from cattr import GenConverter
from attr import attrs

T = TypeVar("T")
c = GenConverter()

@attrs(auto_attribs=True)
class Inner():
    a: int

@attrs(auto_attribs=True)
class Outer(Generic[T]):
    inner: T

initial = Outer[Inner](Inner(1))
raw = c.unstructure(initial, Outer[Inner])
assert raw == {"inner": {"a": 1}}
#      ^^^ getting {'inner': Inner(a=1)} instead
new = c.structure(raw, Outer[Inner])
assert initial == new
```

Tinche:
@Savlik Looks like a bug. I'll see about getting it fixed for the next release!

wakemaster39:
I put some code into #172 for this. 

Its a little rough having to use the `inspect` module to detect the signature, but this allows deeply nested things to work.

MicaelJarniac:
I was having this issue today, and installing the latest (unreleased) version of `cattrs` solved it!

Here are the error messages:

```
  File "/test.py", line 2, in from_bson
    return converter_bson.structure(b, cls)
  File "/venv/lib/python3.8/site-packages/cattr/converters.py", line 294, in structure
    return self._structure_func.dispatch(cl)(obj, cl)
  File "/venv/lib/python3.8/site-packages/cattr/dispatch.py", line 49, in _dispatch
    return self._function_dispatch.dispatch(cl)
  File "/venv/lib/python3.8/site-packages/cattr/dispatch.py", line 126, in dispatch
    return handler(typ)
  File "/venv/lib/python3.8/site-packages/cattr/converters.py", line 733, in gen_structure_attrs_fromdict
    h = make_dict_structure_fn(
  File "/venv/lib/python3.8/site-packages/cattr/gen.py", line 177, in make_dict_structure_fn
    mapping = _generate_mapping(base, mapping)
  File "/venv/lib/python3.8/site-packages/cattr/gen.py", line 156, in _generate_mapping
    return cls(**mapping)
[...]

PR #172, opened 2021-09-28
deeply nested generics

wakemaster39:
@Tinche  Replaces #171. fixes the issue of `Optional[List[Sequence[Set[T]]]]` and `Optional[List[SomeGeneric]]` where the Generic does not have its parameters specified so mapping will be empty. 

* Adds default none type structuring.
* blacked the code base, not sure why it diverged but I can roll it back if you prefer.

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -3,7 +3,7 @@ from collections.abc import MutableSet as AbcMutableSet
 from dataclasses import Field
 from enum import Enum
 from functools import lru_cache
-from typing import Any, Callable, Dict, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Type, TypeVar, Union, Protocol
 
 from attr import Attribute
 from attr import has as attrs_has
@@ -84,6 +84,8 @@ def is_optional(typ):
         and len(typ.__args__) == 2
     )
 
+def is_protocol(typ):
+    return getattr(typ, "_is_protocol", False)
 
 class Converter(object):
     """Converts between structured and unstructured data."""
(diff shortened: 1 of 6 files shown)


Tinche:
Cool! I blacked `bench/` on master in the meantime. Will take a look soon.

src/cattr/converters.py
```diff
@@ -45,7 +45,7 @@
     make_hetero_tuple_unstructure_fn,
     make_iterable_unstructure_fn,
     make_mapping_structure_fn,
-    make_mapping_unstructure_fn,
+    make_mapping_unstructure_fn, _generate_mapping,
```
IceN9ne, line 48: This should probably go on its own line to keep things regular.

Tinche:
I can take a look at this in a week since I'm at a company retreat atm.

wakemaster39:
Sounds good, I keep making some small tweaks and changes as I find more broken things around #170.

The crux of the issue is that unstructuring is based on the field type now and it is not based on the object being passed in. Deeply nested generics, protocols and others are basically hitting the `_unstructure_identity` and getting passed through.

As issues are found new handles will be needed. The possible recursion of the protocol handler that I created calling unstructure again has some smell to it, but I don't see another way right now to unstructure based on the type being passed in. 

This same methodology could also be used for TypeVar, so that there is a fallback handler of instead of hitting the identity. I have not run into it yet in my code but this seems like it could eventually run into issues with a TypeVar in the generic list:

https://github.com/Tinche/cattrs/blob/f377e04f9375d50c4496388f7b0f64d51790561c/src/cattr/gen.py#L292-L302

petergaultney:
I can confirm that does indeed fix it. wow, that was fast. thank you!

petergaultney:
I've done fairly extensive further testing and this seems to have resolved all of our issues with the new GenConverter in `typecats` and the applications it supports.

I would be very happy to see this merged whenever @Tinche and @wakemaster39 agree that it's ready.

Tinche:
So while working on this the fact that `issubclass(int, Protocol) == True` is throwing me off a little. I was not expecting that, heh.

Tinche:
My fix is live on main, including your tests! Let me know if it's good enough!

PR #182, opened 2021-11-11
Fix Literal[...] with more than one item

### LINKED ISSUE #180: Literal[...] with more than one item raises exception
* cattrs version: 1.9.0 (unreleased, installed from GitHub)
* Python version: 3.8.10
* Operating System: Ubuntu running on WSL2 on Windows 10

### Description

The `Literal[...]` type allows for multiple values to be used, like `Literal["one", "two"]`, but `cattr` seems to only support the very first value, and throws an exception when structuring the others.

```py
from typing import Literal, Optional

import attr
import cattr


@attr.s(auto_attribs=True)
class A:
    foo: Optional[Literal["one", "two"]] = None


cattr.structure({}, A)
# A(foo=None)

cattr.structure({"foo": "one"}, A)
# A(foo='one')

cattr.structure({"foo": "two"}, A)
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
#   File "/venv/dev/lib/python3.8/site-packages/cattr/converters.py", line 295, in structure
#     return self._structure_func.dispatch(cl)(obj, cl)
#   File "<cattrs generated structure __main__.A>", line 5, in structure_A
#   File "/venv/dev/lib/python3.8/site-packages/cattr/converters.py", line 526, in _structure_optional
#     return self._structure_func.dispatch(other)(obj, other)
#   File "/venv/dev/lib/python3.8/site-packages/cattr/converters.py", line 404, in _structure_literal
#     raise Exception(f"Literal {type} not equal to {val}")
# Exception: Literal typing.Literal['one', 'two'] not equal to two
```

https://docs.python.org/3/library/typing.html#typing.Literal

This issue has also been reported on https://github.com/python-attrs/cattrs/issues/122#issuecomment-853771560, but since that issue is closed, I decided to open a new one for this.

What I'm currently doing as a work-around is using `Enum`s instead of `Literal`s, since `cattr` has some awesome support for `Enum`s.

MicaelJarniac:
Fixes #180

### DIFF
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -400,8 +400,8 @@ class Converter(object):
 
     @staticmethod
     def _structure_literal(val, type):
-        if val != type.__args__[0]:
-            raise Exception(f"Literal {type} not equal to {val}")
+        if val not in type.__args__:
+            raise Exception(f"{val} not in literal {type}")
         return val
 
     # Attrs classes.
--- a/tests/test_structure_attrs.py
+++ b/tests/test_structure_attrs.py
@@ -166,6 +166,26 @@ def test_structure_literal(converter_cls):
 
 @pytest.mark.skipif(is_py37, reason="Not supported on 3.7")
 @pytest.mark.parametrize("converter_cls", [Converter, GenConverter])
+def test_structure_literal_multiple(converter_cls):
+    """Structuring a class with a literal field works."""
+    from typing import Literal
+
+    converter = converter_cls()
+
+    @define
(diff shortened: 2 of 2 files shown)


MicaelJarniac:
I have absolutely no idea how coverage went down. I even added tests for the things I changed.

Tinche:
LGTM! I wouldn't worry too much about minute changes in coverage, a lot of the tests are Hypothesis tests which aren't strictly deterministic.

Tinche:
Thanks for this!

Issue #184, opened 2021-11-15
Wrong name of parameter in the documentation and tests


jevhen-ponomarenko:
* cattrs version: 1.8.0
* Python version: 3.8.12
* Operating System: Debian GNU/Linux 11 

### Description
I think there is a typo in the documentation and in some parts of the code.

`_cattr_fobid_extra_keys` seems to be a wrong name of the parameter mentioned [in the documentation](https://cattrs.readthedocs.io/en/latest/customizing.html#forbid-extra-keys), instead it should be `_cattrs_fobid_extra_keys`.

### What I Did

```python
import cattr
from cattr.gen import make_dict_structure_fn
from attr import attrib, attributes

@attributes
class TestClass:
    number: int = attrib(default=42)
 c = cattr.GenConverter()
hook = make_dict_structure_fn(TestClass, c, _cattr_forbid_extra_keys=True)
c.register_structure_hook(TestClass, hook)
c.structure({"nummber": 2}, TestClass)

Out[10]: TestClass(number=42)

###############################

hook = make_dict_structure_fn(TestClass, c, _cattrs_forbid_extra_keys=True)
c.register_structure_hook(TestClass, hook)
c.structure({"nummber": 2}, TestClass)

---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-12-8ee276eb47f3> in <module>
      1 c.register_structure_hook(TestClass, hook)
----> 2 c.structure({"nummber": 2}, TestClass)

/opt/venv/lib/python3.8/site-packages/cattr/converters.py in structure(self, obj, cl)
    292         """Convert unstructured Python data structures to structured data."""
    293 
--> 294         return self._structure_func.dispatch(cl)(obj, cl)
    295 
    296     # Classes to Python primitives.

<cattrs generated structure __main__.TestClass-2> in structure_TestClass(o, *_)
      6   unknown_fields = set(o.keys()) - __c_a
      7   if unknown_fields:
----> 8     raise Exception(      9       'Extra fields in constructor for TestClass: ' + ', '.join(unknown_fields)    )     10   return __cl(**res)

Exception: Extra fields in constructor for TestClass: nummber
```

Tinche:
Ah yes, you are correct. We should fix the docs to use `_cattrs_forbid_extra_keys` instead. Mind putting together a PR?

jevhen-ponomarenko:
will do

jevhen-ponomarenko:
[...]

PR #189, opened 2021-11-17
Remove uses of _cattr_forbid_extra_keys in documentation and test

jevhen-ponomarenko:
#184

### DIFF
--- a/docs/customizing.rst
+++ b/docs/customizing.rst
@@ -106,7 +106,7 @@ creating structure hooks with ``make_dict_structure_fn``.
     Traceback (most recent call last):
     ...
     Exception: Extra fields in constructor for TestClass: nummber
-    >>> hook = make_dict_structure_fn(TestClass, c, _cattr_forbid_extra_keys=False)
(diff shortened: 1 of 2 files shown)


Tinche:
Thanks! A little unclear on how the test was passing with the wrong argument though.

Tinche:
> Thanks! A little unclear on how the test was passing with the wrong argument though.

Ah, it's because the argument defaults to false.

You need to format `tests/metadata/test_genconverter.py` with Black for the lint to pass, btw ;)

jevhen-ponomarenko:
> You need to format tests/metadata/test_genconverter.py with Black for the lint to pass, btw ;)

Sorry still having a hard time with installing all of the Python dependencies on Apple Silicon. 

Since the current test suite didn't catch this bug do you think another unit test should be added?

Tinche:
> > You need to format tests/metadata/test_genconverter.py with Black for the lint to pass, btw ;)
> 
> Sorry still having a hard time with installing all of the Python dependencies on Apple Silicon.
> 
> Since the current test suite didn't catch this bug do you think another unit test should be added?

Nah, the test was valid, it was just trying to set the variable to its default value.

PR #198, opened 2021-12-20
Fix test with 32-bit time_t

viccie30:
Check whether 64-bit time_t is supported. If not, make sure to limit
timestamps to 32 bits.

### DIFF
--- a/tests/test_preconf.py
+++ b/tests/test_preconf.py
@@ -103,8 +103,8 @@ def everythings(
         )
     )
     dts = datetimes(
-        min_value=datetime(1900, 1, 1),
-        max_value=datetime(2200, 1, 1),
+        min_value=datetime(1904, 1, 1),
+        max_value=datetime(2038, 1, 1),
         timezones=just(timezone.utc),
     )
     if not allow_datetime_microseconds:


Tinche:
Interesting, can you give a little bit of context for this change? 32-bit Pythons?

viccie30:
Indeed, 32-bit Pythons on systems with 32-bit `time_t`s. On those systems the following line fails for timestamps that don't fit in 32 bits: https://github.com/python-attrs/cattrs/blob/bc9432e606177fe10aa3d2d11e715970c92526be/src/cattr/preconf/msgpack.py#L17.

The Python documentation specifies that `datetime.fromtimestamp` can fail when the timestamp is out-of-range (https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp). 

I ran into this while trying to (privately) package cattrs for Debian, you can see the failure in the CI here: https://salsa.debian.org/viccie30/python-cattr/-/jobs/2295707#L1358

Tinche:
If this is the only problem, I would be OK with just changing all datetimes in that test to be 32-bit (i.e. remove the try/except).

viccie30:
I changed the minimum and maximum to 32 bits like you asked.

Tinche:
Alright, thanks!

Issue #201, opened 2021-12-21
ForwardRef and GenConverter do not seem to work together


aha79:
* cattrs version:
* Python version:
* Operating System:

### Description

When using recursive types (see example below), `structure` with `GenConverter` raises a `StructureHandlerNotFoundError`   with the message  "Unsupported type: ForwardRef('Y'). Register a structure hook for it"
.
The error does not happen with PEP 563 enabled. That means when we add 'from __future__ import annotations' to the program and 
replace `y: typing.Optional['Y']` with `y: typing.Optional[Y]` everything works as expected.

### What I Did

```
import typing
import cattr
import attr

@attr.define(auto_attribs=True)
class X:
    a: int
    b: float
    y: typing.Optional['Y']   #with PEP563 enabled:   y: typing.Optional[Y]   (that does not raise an exception) 

@attr.define(auto_attribs=True)
class Y:
    a: typing.List[int]
    x: X


converter = cattr.GenConverter()

data = { 'a': [1,2,3], 'x': {'a':1, 'b':2.000001, 'y': {'a': [], 'x': {'a': 5, 'b': 3.1, 'y': None}}} }
d = converter.structure( data, Y )  # <---- raises StructureHandlerNotFoundError    Unsupported type: ForwardRef('Y'). Register a structure hook for it
```

Tinche:
Just FYI you don't need `auto_attribs=True` when using `define`, that's the default behavior :)

Tinche:
I was bitten by this lately too. Wrap the entire type in quotes, instead of just `'Y'`:
```
import typing

import attr

import cattr


@attr.define
class X:
    a: int
    b: float
    y: "typing.Optional[Y]"


@attr.define
class Y:
    a: typing.List[int]
    x: X


converter = cattr.GenConverter()

data = {
    "a": [1, 2, 3],
    "x": {
        "a": 1,
        "b": 2.000001,
        "y": {"a": [], "x": {"a": 5, "b": 3.1, "y": None}},
    },
}
d = converter.structure(data, Y)
```

It has to do with how `typing.get_type_hints` works.

aha79:
Okay, I see. I confirm, this works.

However now I try to get this one-line recursive type

    YY = typing.List[ typing.Tuple[typing.Optional[YY],int] ]

to work.

Unfortunately, this generates a "NameError" even if PEP563 is enabled (I know this is not a cattrs issue. perhaps this is not supported by PEP563). 

[...]

PR #215, opened 2022-01-27
Support ForwardRef

### LINKED ISSUE #206: TypeError when registering ForwardRefs in python 3.10.2
* cattrs version: 1.10.0
* Python version: 3.10.2
* Operating System: Ubuntu 20.04.3

### Description
I have a converter that registers a `ForwardRef`. In python 3.6.x through 3.10.1, this has worked as intended. As of python 3.10.2, this is now throwing a `TypeError`:
```python
TypeError: Invalid first argument to `register()`. ForwardRef('MyClass') is not a class.
```

Strangely, I don't see anything in the [python 3.10.2 changelog](https://docs.python.org/3.10/whatsnew/changelog.html#python-3-10-2-final) that would indicate changed behavior with `ForwardRef` or `register()`.

### What I Did

Minimal example:
```python
from cattr import Converter
from typing import ForwardRef

class MyClass:
    ...


converter = Converter()
    ForwardRef('MyClass'), (lambda obj, _: converter.structure(obj, MyClass))
)
```

Traceback:
```python
TypeError                                 Traceback (most recent call last)
Input In [17], in <module>
----> 1 converter.register_structure_hook(
      2     ForwardRef("MyClass"), (lambda obj, _: converter.structure(obj, MyClass))
      3 )

File ~/.virtualenvs/rc-310/lib/python3.10/site-packages/cattr/converters.py:269, in Converter.register_structure_hook(self, cl, func)
    267     self._structure_func.clear_cache()
    268 else:
--> 269     self._structure_func.register_cls_list([(cl, func)])

File ~/.virtualenvs/rc-310/lib/python3.10/site-packages/cattr/dispatch.py:57, in MultiStrategyDispatch.register_cls_list(self, cls_and_handler, direct)
     55         self._direct_dispatch[cls] = handler
     56     else:
---> 57         self._single_dispatch.register(cls, handler)
     58         self.clear_direct()
     59 self.dispatch.cache_clear()

File ~/.pyenv/versions/3.10.2/lib/python3.10/functools.py:856, in singledispatch.<locals>.register(cls, func)
    854 else:
    855     if func is not None:
--> 856         raise TypeError(
    857             f"Invalid first argument to `register()`. "
    858             f"{cls!r} is not a class."
    859         )
    860     ann = getattr(cls, '__annotations__', {})
    861     if not ann:

TypeError: Invalid first argument to `register()`. ForwardRef('MyClass') is not a class.
```

### Context
It's possible this isn't something that needs to be fixed in `cattrs`, and I'm just not using it correctly. Here's a little context for what I'm trying to do, in case anyone has advice for a better way to accomplish the same thing.

I have a class that contains a list of instances of itself, so that attribute is typed as a forward reference. As for "why would you want to do that," this is in a model based on the [requests.Response](https://docs.python-requests.org/en/latest/api/#requests.Response) class, which contains a list of `Response` objects representing redirect history. I then basically need to tell `cattrs` to resolve that forward reference using an unstructure hook. Here's a minimal example:
```python
import attr
from typing import List

@attr.define()
class MyClass:
    history: List['MyClass'] = attr.field(factory=list)
```

Full class: https://github.com/reclosedev/requests-cache/blob/a29b85c3df40f874f80d06c949463b9532dada40/requests_cache/models/response.py#L35
Full converter: https://github.com/reclosedev/requests-cache/blob/a29b85c3df40f874f80d06c949463b9532dada40/requests_cache/serializers/cattrs.py#L68-L72
Related issue: https://github.com/reclosedev/requests-cache/issues/501

aha79:
This adds ForwardRef support to cattr (See #201). 

- ForwardRefs in classes 
   ```
   @dataclass
   class Test:
       x: List["XType"]

   XType = int
   ```
- Recursive classes
   ```
   @dataclass
   class Node:
       data: Dict
       left: Optional["Node"]
       right: Optional["Node"]
  ```

- Recursive type aliases. (Note: Even with [PEP 563](https://www.python.org/dev/peps/pep-0563/) we need ForwardRefs here)
   ```
   # Tree-like structure
   Node = Tuple[ Dict, Optional["Node"], Optional["Node"] ]

   class UserClass:
       some_tree_data: Node
   ```
- Recursive type aliases, across modules. (Needs explicit ForwardRef resolution, though. This is a Python limitation)
   ```
   # module.py
   Node = Tuple[ Dict, Optional["Node"], Optional["Node"] ]
   cattr.resolve_types(Node, globals(), locals())

   # main.py
   from module import Node
   class UserClass:
       some_tree_data: Node
   ```
- Manual registration (this worked before, but is still supported.) This PR also fixes #206)
   ```
   Node = Tuple[ Dict, Optional["Node"], Optional["Node"] ]

   c.register_structure_hook(
       ForwardRef("Node"), lambda obj, _: c.structure(obj, Node)
   )
   c.register_unstructure_hook(
       ForwardRef("Node"), lambda obj: c.unstructure(obj, Node)
   )
  ```

 
Specifically, the PR does

- Add hooks for ForwardRef (this equires ForwardRefs to be in resolved state, otherwise an error is generated)
- Ensure that ForwardRef are resolved for registered types, whenever it is possible to resolve them without context information. (Essentially this is done by a call `get_type_hints` on the type containing the ForwardRef).

   For this a new function `resolve_types` was added which is systematically used. It basically does the same thing as `attrs.resolve_types` but works for all types  (attrs classes, dataclasses and type aliases).
- ForwardRefs are now registered with direct dispatch and no longer singledispatch. This is to work around #206.

### DIFF
--- a/src/cattr/__init__.py
+++ b/src/cattr/__init__.py
@@ -1,5 +1,6 @@
 from .converters import Converter, GenConverter, UnstructureStrategy
 from .gen import override
+from ._compat import resolve_types
 
 __all__ = (
     "global_converter",
@@ -11,6 +12,7 @@ __all__ = (
     "Converter",
     "GenConverter",
     "override",
+    "resolve_types",
 )
 
 
--- a/src/cattr/_compat.py
+++ b/src/cattr/_compat.py
@@ -1,8 +1,9 @@
 import sys
 from dataclasses import MISSING
 from dataclasses import fields as dataclass_fields
-from dataclasses import is_dataclass
-from typing import Any, Dict, FrozenSet, List
+from dataclasses import is_dataclass, make_dataclass
+from dataclasses import Field as DataclassField
+from typing import Any, Dict, FrozenSet, List, Optional
 from typing import Mapping as TypingMapping
 from typing import MutableMapping as TypingMutableMapping
 from typing import MutableSequence as TypingMutableSequence
@@ -13,7 +14,8 @@ from typing import Tuple, get_type_hints
 
 from attr import NOTHING, Attribute, Factory
 from attr import fields as attrs_fields
-from attr import resolve_types
+from attr import resolve_types as attrs_resolve_types
+from attr import has as attrs_has
 
 version_info = sys.version_info[0:3]
 is_py37 = version_info[:2] == (3, 7)
@@ -373,3 +375,81 @@ else:
 
 def is_generic_attrs(type):
     return is_generic(type) and has(type.__origin__)
+
+
+def resolve_types(
+    cls: Any,
+    globalns: Optional[Dict[str, Any]] = None,
+    localns: Optional[Dict[str, Any]] = None,
+):
+    """
+    More generic version of `attrs.resolve_types`.
+
+    While `attrs.resolve_types` resolves ForwardRefs
+    only for for the fields of a `attrs` classes (and
+    fails otherwise), this `resolve_types` also
+    supports dataclasses and type aliases.
+
+    Even though often ForwardRefs outside of classes as e.g.
+    in type aliases can generally not be resolved automatically
+    (i.e. without explicit `globalns`, and `localns` context),
+    this is indeed sometimes possible and supported by Python.
+    This is for instance the case if the (internal) `module`
+    parameter of `ForwardRef` is set or we are dealing with
+    ForwardRefs in `TypedDict` or `NewType` types.
+    There may also be additions to typing.py module that there
+    will be more non-class types where ForwardRefs can automatically
+    be resolved.
+
+    See
+        https://bugs.python.org/issue41249
+        https://bugs.python.org/issue46369
+        https://bugs.python.org/issue46373
+    """
+    allfields: List[Union[Attribute, DataclassField]] = []
+
+    if attrs_has(cls):
+        try:
+            attrs_resolve_types(cls, globalns, localns)
+        except NameError:
+            # ignore if ForwardRef cannot be resolved.
+            # We still want to allow manual registration of
+            # ForwardRefs (which will work with unevaluated ForwardRefs)
+            pass
+        allfields = fields(cls)
+    else:
+        if not is_dataclass(cls):
+            # we cannot call get_type_hints on type aliases
+            # directly, so put it in a field of a helper
+            # dataclass.
+            cls = make_dataclass("_resolve_helper", [("test", cls)])
+
+            # prevent resolving from cls.__module__ (which is what
+            # get_type_hints does if localns/globalns == None), as
+            # it would not be correct here.
+            # See: https://stackoverflow.com/questions/49457441
+            if globalns is None:
+                globalns = {}
+            if localns is None:
+                localns = {}
+        else:
+            allfields = dataclass_fields(cls)
+
+        try:
+            type_hints = get_type_hints(cls, globalns, localns)
+            for field in allfields:
+                field.type = type_hints.get(field.name, field.type)
+        except NameError:
+            pass
+    if not is_py39_plus:
+        # 3.8 and before did not recursively resolve ForwardRefs
+        # (likely a Python bug). Hence with PEP 563 (where all type
+        # annotations are initially treated as ForwardRefs) we
+        # need twice evaluation to properly resolve explicit ForwardRefs
+        fieldlist = [(field.name, field.type) for field in allfields]
+        cls2 = make_dataclass("_resolve_helper2", fieldlist)
+        cls2.__module__ = cls.__module__
+        try:
+            get_type_hints(cls2, globalns, localns)
+        except NameError:
+            pass
--- a/src/cattr/converters.py
+++ b/src/cattr/converters.py
@@ -3,11 +3,11 @@ from collections.abc import MutableSet as AbcMutableSet
 from dataclasses import Field
 from enum import Enum
 from functools import lru_cache
-from typing import Any, Callable, Dict, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, Dict, ForwardRef, Optional
+from typing import Tuple, Type, TypeVar, Union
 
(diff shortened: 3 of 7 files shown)


Tinche:
Thanks for putting in work on this.

The hooks are obviously fine. The issue is that this introduces `resolve_types` for dataclasses, and tbh that's out of scope for cattrs. That should be implemented in Python itself or another library that's willing to take on the maintenance burden of it.

aha79:
Hmm, but what is the alternative? Not supporting dataclasses would not be very useful - at least not for me.
Hooks without any type resolving before will just generate an error message.

In fact, you already do resolve dataclasses in cattr. It is the `get_type_hints` call inside `adapted_fields`. Even though this is enough to handle PEP563, one needs more symmetric handling of attrs and dataclasses to get all tests to pass. That is why I introduced `resolve_types`, in the same spirit as `has` and `fields`in _compat.py.

One could simplify the new `resolve_types` by removing the part under `if not is_py39_plus:`, though. This would not really hurt as it only affects older Python versions.

aha79:
I understand your concern regarding mantainability. However I would argue having a more symmetric way of handling dataclasses and attrs classes is cleaner and needs less maintenance effort. 

For instance: One can get rid of `get_type_hints` call in `adapted_fields` (which probably should not really be in there) and the calls to `attrs.resolve_types` scattered around cattrs (and some more clean ups). 
But that would be a refactoring on its own and I wanted to keep the list of changes for this PR small.

aha79:
Could it be that in https://github.com/python-attrs/cattrs/commit/a4282724a47df8ac9f5a9c8069dc49160cae152d cattr/gen.py and cattrs/gen.py are accentially swapped?

Tinche:
You make a couple of good points, so let's discuss further.

(The cattr -> cattrs move isn't accidental, it's on purpose. I think all new code will be going into the cattrs namespace, and existing code will be migrating there over time. The old packages will still have aliases so no backwards compatibility break.)

The `resolve_types` as you've implemented it overwrites the `type` field for dataclasses. I have no idea if this is a supported use case, and that's why cattrs is careful to never do it. Hence the adapted fields approach. If in the next version Python decides to make this impossible by making this field immutable or something, I don't want this to be a cattrs problem.

On a higher level, here's the philosophy of cattrs: attrs is simply a better library than dataclasses, and one of its advantages is that the maintainers of attrs (I'm one of them) choose to implement and maintain `resolve_types`. The maintainers of dataclasses choose not to - instead the interface provided is `get_type_hints`, and the users get to deal with that. The answer here isn't cattrs patching up this missing support in dataclasses, it's the users of dataclasses (you) switching to attrs or putting pressure on the maintainers of dataclasses to implement this.

Now that you see where I'm coming from, you say you're forced to use dataclasses and I'm sympathetic to this use case. So let's discuss to see what we can accomplish without cattrs implementing `resolve_types` for dataclasses in this form.

One of the options is you implementing `resolve_types` for dataclasses in your code base. So you get to maintain it instead ;) Would this work for you? Another option would be improving some of the cattrs APIs to be more customizable, which would be useful for other things too.

Also, didn't you mention ForwardRefs can resolve themselves in newer Python versions? Presumably via `get_typing_hints`?

aha79:
The primary reason to implement `resolve_types` for all types is that it is needed inside cattr for `ForwardRef` resolution and not to provide it to the user (that would be just a side effect). 

Perhaps let me first expand on what functionality is needed:

- You are right that Python (only) provides the `get_type_hints` function for type resolution. In fact this is all we need. The thing that makes it complicated is that we cannot just call `get_type_hints` on the ForwardRef and expect it to work. 
- Python/`get_type_hints`, under the hood, does some complicated things to find out the target module of the ForwardRef. Some types know their module and `get_type_hints` will resolve the ForwardRefs in child annotations. 
For instance `get_type_hints` called on a dataclass with a field of type `Optional['X']` will resolve the 'X' in the same module scope as the dataclass was defined, no matter where the call to `get_type_hints` is made. 
 
- Types that know their own module are: `ForwardRefs` with `module` field set (only used in special cases), classes (including attrs classes & dataclasses), NewType, TypedDict, NamedTuple, and perhaps more I overlooked [1].  
- So to resolve ForwardRefs we need to call `get_type_hints` on some __parent__ type of the ForwardRef.  When this parent type knows its module `get_type_hints` will resolve the child ForwardRefs.
- Concluding, since it is beyond the scope of a library such as cattr to know which typing.py types carry module scope and which not, the simplest solution is to just call `get_type_hints` on every type that is supposed to get (un)structured. That makes sure that all child-ForwardRefs get resolved when there is opportunity for them to get resolved. 


The original idea was to concentrate all `get_type_hints` business logic calls into one function which is called in the hook factory and takes care of resolving ForwardRefs and (the closely related) PEP563 delayed annotations.

So now let me try to address your specific concerns: I see several options:
   1.  Remove overwriting of `type` fields of dataclasses in `resolve_types`. You make a valid point of not overwriting the `type` field of dataclass `Field`. This is not good. Removing it will just introduce failing tests with `Converter`. But `GenConverter` is still fine. If ForwardRefs would only be supported with GenConverter this is ok for me. 
   2.  Make `resolve_types`  return a value. After `get_type_hints` do not write to `type` field, but convert dataclasses to attrs classes. This is essentially the functionality that is done in `adapted-fields`, and not only return a list of Attribute but put these in a attrs class and return that. (This and corresponding clean-ups could also be done after 1 and we could would not strictly need that)
   3. The `resolve_types` does not need to be exported from cattr.
   4. `resolve_types`  could receive a better name 

So if we do 1 & 3 and possibly 4, the resolve_types would have the dataclasses handling removed and not be exported. It would not generate an error when called for datacasses and work for type aliases as well.


[1]: Even though these types are not (yet) supported by cattrs, the user might still register a hook for them and expect child ForwardRefs to work.

aha79:
> Another option would be improving some of the cattrs APIs to be more customizable, which would be useful for other things too.

Could you elaborate on that?

Issue #217, opened 2022-02-02
[Bug] GenConverter: Wrong resolving / missing __parameters__ attribute when inheriting something from `typing.*` 


raabf:
* cattrs version: 1.10.0
* Python version: 3.7
* Operating System: Linux 5.16.1-1 OpenSUSE Thumbleweed

There is an exception `AttributeError: type object 'CLASS' has no attribute '__parameters__'` when you try to structure an attrs Class which inherts `typing.CLASS`

## Reproduce

Have a look at this example where `CLASS` is `Hashable`:

```python
import attr
import typing
import cattrs
import collections

converter = cattrs.GenConverter()

@attr.define
class A(collections.abc.Hashable):  # typing.Hashable should be preferred over collections.abc.Hashable
    attr1: str = ''
    attr2: int = 0

    def __hash__(self) -> int:
        return hash(self.attr2)

@attr.define
class B(typing.Hashable):
    attr1: str = ''
    attr2: int = 0

    def __hash__(self) -> int:
        return hash(self.attr2)


def main() -> None:
    converter.structure(dict(attr1='hello', attr2=42), A)  # OK
    converter.structure(dict(attr1='hello', attr2=42), B)  # AttributeError: type object 'Hashable' has no attribute '__parameters__'

if __name__ == '__main__':
    main()
```

Output
```
Traceback (most recent call last):
  File "/home/raabf/sl/demo.py", line 40, in <module>
    main()
  File "/home/raabf/sl/demo.py", line 37, in main
    converter.structure(dict(attr1='hello', attr2=42), B)  # AttributeError: type object 'Hashable' has no attribute '__parameters__'
  File "/home/raabf/.pyenv/versions/sl37/lib/python3.7/site-packages/cattr/converters.py", line 300, in structure
    return self._structure_func.dispatch(cl)(obj, cl)
  File "/home/raabf/.pyenv/versions/sl37/lib/python3.7/site-packages/cattr/dispatch.py", line 49, in _dispatch
    return self._function_dispatch.dispatch(cl)
  File "/home/raabf/.pyenv/versions/sl37/lib/python3.7/site-packages/cattr/dispatch.py", line 126, in dispatch
    return handler(typ)
  File "/home/raabf/.pyenv/versions/sl37/lib/python3.7/site-packages/cattr/converters.py", line 745, in gen_structure_attrs_fromdict
    **attrib_overrides,
[...]

PR #221, opened 2022-02-03
Fix  __parameters__ access in gen._generate_mapping

raabf:
There are Generic types in the typing modules
from which you can inherit in your own classes
which do not have an __parameters__ attribute,
such classes are now ignored making
gen._generate_mapping effectively a no-op
in case the class do not have an __parameters__
attribute.

As https://github.com/ilevkivskyi/typing_inspect/blob/8f6aa2075ba448ab322def454137e7c59b9b302d/typing_inspect.py#L405
is showing there are also cases where __parameters__
could be None, so I test for both cases, that it
is None or that it does not exist.

See Also:
https://github.com/python-attrs/cattrs/issues/217

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,6 +12,9 @@ History
 * Fix structuring bare ``typing.Tuple`` on Pythons lower than 3.9.
   (`#218 https://github.com/python-attrs/cattrs/issues/218`_)
 
+* Fix a wrong ``AttributeError`` of an missing ``__parameters__`` attribute. This could happen 
+  when inheriting certain generic classes – for example ``typing.*`` classes are affected. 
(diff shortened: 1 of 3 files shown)


Tinche:
Alright, nice and easy. Add a changelog entry, make sure tests pass and in it goes ;)

Tinche:
Looks good. There's a conflict since I merged something else in prior so please resolve that and I'll merge. I'll probably edit your changelog entry for brevity.

Thanks!

raabf:
Cool!
I have shortened the history entry a bit, feel free to edit further.
I have rebased the branch on current main.

Tinche:
Test failures I believe are unrelated, and I fixed them yesterday. Merging this in, thanks!

PR #224, opened 2022-02-05
Fix issue links in HISTORY

bibajz:
Sorry, I messed up the links to issues resolved - they display incorrectly and look like in the screenshot below:

Screenshot 2022-02-05 at 11-10-27 python-attrs cattrs Complex custom class converters for attrs


Tinche:
Actually I messed those up when editing yours, but thanks ;)

PR #225, opened 2022-02-07
For `forbid_extra_keys` raise custom `ForbiddenExtraKeyError`

raabf:
When `forbid_extra_keys` is actived, I would like to be able to try-except that specific Problem (here `ForbiddenExtraKeyError`) instead of any `Exception`.

The unit tests also gain a little more percition through that change.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -18,7 +18,8 @@ History
   (`#231 <https://github.com/python-attrs/cattrs/pull/231>`_)
 * Fix unstructuring all tuples - unannotated, variable-length, homogenous and heterogenous - to `list`.
   (`#226 <https://github.com/python-attrs/cattrs/issues/226>`_)
-
+* For ``forbid_extra_keys`` raise custom ``ForbiddenExtraKeyError`` instead of generic ``Exception``.
+  (`#255 <https://github.com/python-attrs/cattrs/pull/225>`_)
 
 1.10.0 (2022-01-04)
 -------------------
--- a/docs/customizing.rst
+++ b/docs/customizing.rst
@@ -108,7 +108,7 @@ creating structure hooks with ``make_dict_structure_fn``.
     >>> c.structure({"nummber": 2}, TestClass)
     Traceback (most recent call last):
     ...
-    Exception: Extra fields in constructor for TestClass: nummber
+    ForbiddenExtraKeyError: Extra fields in constructor for TestClass: nummber
     >>> hook = make_dict_structure_fn(TestClass, c, _cattrs_forbid_extra_keys=False)
     >>> c.register_structure_hook(TestClass, hook)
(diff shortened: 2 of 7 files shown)


Tinche:
I think this is a good idea but it should wait until after the extended validation branch is merged, which is blocked by other things temporarily.

raabf:
Sure, order the MR as it fits best. :smile:

Tinche:
Hello, I'm preparing a new release and let's get this merged!

So please rebase, and I would ask for a change: add a field to the exception (`extra_keys`) that carries the list of extra keys. This will make it easier to introspect these exceptions programmatically.

raabf:
> add a field to the exception (extra_keys) that carries the list of extra keys. This will make it easier to introspect these exceptions programmatically.

Yea that is nice – I have added this and a reference to class to which the extra keys should be used (i.e. to which class constructor the extra keys were passed). I also extended unit test to check that attributes.

However, there is a small problem, since `ForbiddenExtraKeysError` can conflict with `ClassValidationError`. I assumed that an  `ForbiddenExtraKeysError` means that the Class Validation failed. That means if `ClassValidationError` is used, I add `ForbiddenExtraKeysError` to the `errors` list which is then passed to the `ClassValidationError`. This is more or less anyway implicitly done, because if `ForbiddenExtraKeysError` in a nested class is raised it will be catched anyway by its upper class because you have added ` except Exception as e:` to every attribute. See this commit where I have implemented this behaviour https://github.com/python-attrs/cattrs/pull/225/commits/fd5a9b842da2185b3d04046c1fa38aa533e209da . I hope this is what you have intended.

raabf:
So that it is easy to understand this is an example generated structure function (unknows_fields are added to `errors` list):

```python
def structure_A(o, _, *, __cl=__cl, __c_cve=__c_cve, __c_structure_c=__c_structure_c, __c_type_c=__c_type_c, __c_structure_a=__c_structure_a):
  res = {}
  errors = []
  try:
    res['c'] = __c_structure_c(o['c'], __c_type_c)
  except Exception as e:
    e.__note__ = 'Structuring class test_forbid_extra_keys_nested_override.<locals>.A @ attribute c'
    errors.append(e)
  if 'a' in o:
    try:
      res['a'] = __c_structure_a(o['a'])
    except Exception as e:
      e.__note__ = 'Structuring class test_forbid_extra_keys_nested_override.<locals>.A @ attribute a'
      errors.append(e)
  unknown_fields = set(o.keys()) - __c_a
  if unknown_fields:
    errors.append(__c_feke('', __cl, unknown_fields)) 
  if errors: raise __c_cve('While structuring A', errors, __cl)
  try:
    return __cl(
      **res,
    )
  except Exception as exc: raise __c_cve('While structuring A', [exc], __cl)
```

Tinche:
Yep, that looks great. Thanks a lot!

PR #229, opened 2022-02-13
fix test failure on Windows

brakhane:
On Windows, `datetime.fromtimestamp` only seems to support positive values, cause test_msgpack in test_preconf to fail with OSError.


brakhane:
> Oh, interesting. Let's just make it always generate datetimes from 1970 then.

OK, I'll just update this PR

Tinche:
Thanks!

PR #252, opened 2022-04-10
Make `cattrs` and `cattr.errors` namespaces contain all error types from `cattrs.errors`

jap:
Without this, trying to do something along the lines of
```python
import cattr.errors

try:
  ...
except cattr.errors.BaseValidationError as exc:
  ...
```
fails miserably because it was only importing `StructureHandlerNotFoundError` from `cattrs.errors`

Besides that, importing `cattrs` (with s) causes `cattrs.errors` to be imported from `cattr.errors` which then causes the same problem.

`import cattrs.errors` also has this issue but Python's import machinery is too complicated for me to understand or explain why.

The only way to get a reachable `BaseValidationError` is by doing
```python
from cattrs.errors import BaseValidationError
```
which would cause code review remarks during the day job where we're trying to get rid of "from-imports".


Tinche:
Hi,

`cattr` is kind of a deprecated namespace, it's not going away but new stuff is only going to be added to `cattrs` to gently encourage migration. I have very recently moved the actual errors to `cattrs.errors`, so my first question would be if the `main` branch still has the problem you're experiencing?

jap:
It is, with your main branch checked out:
```
lil-bob:cattrs spaans$ pip install --editable .
Obtaining file:///Users/spaans/src/cattrs
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Collecting exceptiongroup
  Using cached exceptiongroup-1.0.0rc2-py3-none-any.whl (10 kB)
Collecting attrs>=20
  Using cached attrs-21.4.0-py2.py3-none-any.whl (60 kB)
Building wheels for collected packages: cattrs
  Building editable for cattrs (pyproject.toml) ... done
  Created wheel for cattrs: filename=cattrs-22.2.0.dev0-py3-none-any.whl size=4915 sha256=f4138de6c395a84340443b1d66191eb3fba37cc1e6032639ae4580eeefb5ac6a
  Stored in directory: /private/var/folders/hs/y6mmqk9d4p97tr05d11crdb80000gn/T/pip-ephem-wheel-cache-grv9nq2h/wheels/cb/07/86/7e063245dba45861348d66e20e3ff2da1b3ebd669531de559d
Successfully built cattrs
Installing collected packages: exceptiongroup, attrs, cattrs
Successfully installed attrs-21.4.0 cattrs-22.2.0.dev0 exceptiongroup-1.0.0rc2
lil-bob:cattrs spaans$ python3
Python 3.10.3 (main, Mar 29 2022, 22:28:06) [Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cattrs.errors
>>> dir(cattrs.errors)
['StructureHandlerNotFoundError', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__']
```

jap:
and with my branch:
```
lil-bob:cattrs spaans$ git checkout origin/error-import 
Previous HEAD position was 9e740a8 Tweak test
HEAD is now at 079b6f7 Ensure all error types are reachable through module imports
lil-bob:cattrs spaans$ python3
Python 3.10.3 (main, Mar 29 2022, 22:28:06) [Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cattr.errors
>>> dir(cattrs.errors)
  File "<stdin>", line 1, in <module>
NameError: name 'cattrs' is not defined
>>> import cattrs.errors
>>> dir(cattrs.errors)
['BaseValidationError', 'ClassValidationError', 'ForbiddenExtraKeysError', 'IterableValidationError', 'StructureHandlerNotFoundError', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__']
```

jap:
I think the `from cattr import ..., errors, ...` in `cattrs.__init__` is to blame, so it may be easier to fix by just letting that import from its own `errors` module, but then there still is the issue that `cattr.errors` only imports one class instead of all for people using the legacy namespace.

jap:
I put an additional commit in https://github.com/jap/cattrs/tree/error-import2 which takes the above into account, and actually exposes the errors on the cattrs namespace directly, as implied by https://cattrs.readthedocs.io/en/latest/validation.html (which refers to `cattrs.ClassValidationError` for example, which currently really doesn't exist.)

Tinche:
Sorry, I might have posted my reply too early. Can you try with 792a5f4cc1a7e0926f7346406ad693cd55c927b4?

This is what I'm seeing:
```shell
$ python -c "import cattrs.errors; print(dir(cattrs.errors))"
['BaseValidationError', 'ClassValidationError', 'ExceptionGroup', 'ForbiddenExtraKeysError', 'IterableValidationError', 'Optional', 'Set', 'StructureHandlerNotFoundError', 'Type', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__']
```

jap:
Confirmed - your latest version fixes the original issue for me.
Note that the `cattr.errors` namespace is still like it was before, but maybe that is not an issue given that it is deprecated; however, `cattrs` itself doesn't contain the error classes like the validation page implies.

Tinche:
Agreed, would be great if you could open a new PR (or repurpose this one) to just re-export the exceptions from `cattrs.__init__` ;)

jap:
I'll make some time for that this week!

jap:
Rebased against your main branch and made things work.

Tinche (approved):

Tinche:
Sorry, I forgot about this for a while. Can you add a small HISTORY entry and I can merge?

jap:
Sure, done!

Tinche:
Thanks, I merged this into `main` manually manually!

Issue #111, opened 2021-01-08
Mypy error on ctructure


gleb-chipiga:
* cattrs version: 1.1.2
* Python version: 3.8.6
* Operating System: Linux x86-64

### Description

Run mypy on first example from readme and get error.

### What I Did

```python
import cattr
from typing import Tuple

cattr.structure([1.0, 2, "3"], Tuple[int, int, int])

```

```bash
$ mypy example.py
example.py:4: error: Argument 2 has incompatible type "object"; expected "Type[<nothing>]"
Found 1 error in 1 file (checked 1 source file)
```

Tinche:
Ah, I need to tweak the type annotation (turns out `Type[C]` basically only works for classes, not abstract types like `Tuple` or `Optional`). Will do in the next couple of days.

Tinche:
Note that I have to change this type annotation to `Any`, since there's no better way to type check this atm.

Tinche:
I actually needed to add some overloads. There's no way to correctly check `cattr.structure([1.0, 2, "3"], Tuple[int, int, int])` with current Mypy as far as I'm aware.

Tinche:
We're waiting on this: cattr.structure([1.0, 2, "3"], Tuple[int, int, int])

Tinche:
We're waiting on this: https://github.com/python/mypy/issues/9773

PR #259, opened 2022-04-26
Improve method annotations

layday:
When registering an unstructure hook the type checker is not able to
infer the type of `T`; Pyright considers its type to be unknown
(eq. to object).  When attempting to call a method on `T` a type error
is produced.

### DIFF
--- a/src/cattrs/converters.py
+++ b/src/cattrs/converters.py
@@ -184,7 +184,7 @@ class BaseConverter:
         # Unions are instances now, not classes. We use different registries.
         self._union_struct_registry: Dict[Any, Callable[[Any, Type[T]], T]] = {}
 
-    def unstructure(self, obj: Any, unstructure_as=None) -> Any:
+    def unstructure(self, obj: Any, unstructure_as: Optional[type] = None) -> Any:
         return self._unstructure_func.dispatch(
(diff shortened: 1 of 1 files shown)


Tinche:
Some special types like unions and optionals aren't covered by type[T], so these cannot be annotated like this. They used to be, a few versions back, but I had to revert it. We'd need something like https://github.com/python/mypy/issues/9773 to get accepted and merged first.

Tinche:
See also https://github.com/python-attrs/cattrs/issues/111

layday:
Thanks for explaining. Would it be possible to use Any instead of a type variable to avoid having to silence the resultant type errors?

Sent with [ProtonMail](https://protonmail.com/) secure email.

------- Original Message -------

layday:
In `register_unstructure_hook`, replacing the stand-alone type variable. It would also be helpful to annotate the `unstructure_as` parameter of `unstructure` with `Optional[Any]`.

Tinche:
I tweaked the annotations. I don't think `Optional[Any]` makes sense though (`Any` already contains `None`), so I just annotated it as `Any`.

layday:
Great, thanks!

> I don't think `Optional[Any]` makes sense though (`Any` already contains `None`)

Well, it conveys intent, even if `Optional[Any] < Any`.

PR #283, opened 2022-07-04
Raise on input/tuple-args length mis-match

### LINKED ISSUE #57: Tuple[A, B] can be structured from a list with less than two values
* cattrs version: 0.9.0
* Python version: 3.7.2
* Operating System: Win 10 x64

### Description

When structuring a list using the type `Tuple[float, float]`, cattrs will happily construct a tuple from a list with less than two values. I was expecting an error instead. Is this intentional?

### What I Did

```
In [17]: from typing import Tuple

In [18]: cattr.structure([1,2], Tuple[float, float])
Out[18]: (1.0, 2.0)

In [19]: cattr.structure([1], Tuple[float, float])
Out[19]: (1.0,)

In [20]: cattr.structure([], Tuple[float, float])
Out[20]: ()
```

EpicWink:
When structuring a heterogeneous tuple, if the input has too few or many elements, this PR now raises an error. The error is a `ValueError` in non-detailed validation.

Two existing tuple-structuring tests had to be fixed (in the zero-element case).

Fixes #57

### DIFF
--- a/src/cattrs/converters.py
+++ b/src/cattrs/converters.py
@@ -3,7 +3,18 @@ from collections.abc import MutableSet as AbcMutableSet
 from dataclasses import Field
 from enum import Enum
 from functools import lru_cache
-from typing import Any, Callable, Dict, Optional, Tuple, Type, TypeVar, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 from attr import Attribute
 from attr import has as attrs_has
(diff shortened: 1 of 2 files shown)


Tinche:
Hello, thanks for contributing.

Unfortunately this PR makes the happy path almost twice slower.

```
$ pyperf timeit -g -s 'from cattrs import structure; t = tuple[int, str];v = [1, 2]' 'structure(v, t)'
old: Mean +- std dev: 1.50 us +- 0.04 us
new: Mean +- std dev: 2.94 us +- 0.14 us

# Detailed validation false:
old: Mean +- std dev: 1.39 us +- 0.06 us
new: Mean +- std dev: 1.95 us +- 0.09 us
```

You're overcomplicating it with the `_zip_strict` thing. One length comparison after the transformation should be enough, right?

EpicWink:
> One length comparison after the transformation should be enough, right?

The one case that that doesn't solve is when `obj` is a un-sized operable with Moreno values than the tuple type. If you're fine with that trade-off, then I'll make the change

Tinche:
What is `an un-sized operable with Moreno values`?

EpicWink:
> What is `an un-sized operable with Moreno values`?

Ah, autocorrect. I've updated my previous message

EpicWink:
Doing the post-check comparing the result length with the input (latest commit) gives benchmark results:
* Detailed validation:
  * Before: 1.96 ± 0.14 μs
  * After: 2.10 ± 0.15 μs
* Not-detailed validation:
  * Before: 1.83 ± 0.09 μs
  * After: 1.98 ± 0.16 μs

Tinche:
I'll merge this in now. I might work on it a little more myself. Thanks!

Issue #277, opened 2022-06-20
Syntax Error in generated code


phiresky:
* cattrs version: current master
* Python version: 3.10
* Operating System: Linux

### The error

```
 File ".venv/lib/python3.10/site-packages/cattrs/dispatch.py", line 122, in dispatch                                                
   return handler(typ)                                                                                                                                       
 File ".venv/lib/python3.10/site-packages/cattrs/converters.py", line 829, in gen_structure_mapping                                      
   h = make_mapping_structure_fn(                                                                                                                                         
 File ".venv/lib/python3.10/site-packages/cattrs/gen.py", line 726, in make_mapping_structure_fn
   eval(compile(script, "", "exec"), globs)                                                                                                                                
 File "<string>", line 17                                                                                                                                                 
   raise IterableValidationError('While structuring dict[typing.Union[typing.Literal['ca', 'pro', 'fa', 'car', 'ing_emb'], typing.Literal['wgr', 'cwg', 'fcc', 'wcc']], statistics.FrequenciesInfo]', errors, __cattr_mapping_cl)

    | SyntaxError: invalid syntax. Perhaps you forgot a comma?                                                                                                                 
```

The offending code is here gen :

![image](https://user-images.githubusercontent.com/2303841/174582658-085a05de-b325-4afc-abb7-4aa07e02bb9e.png)

It seems to generate code that assumes that the repr of the offending type does not include '. `'` is not escaped.

Tinche:
Thanks, I'm aware of this (there's another, similar issue), it's on my radar for the next version. Just out of curiosity, if you change the line to

```
    raise IterableValidatoinError("While structuring {cl!r}", errors, __cattr_mapping_cl)
```
i.e. single quotes to double quotes, does it fix it?

phiresky:
Yes, that does seem to fix it. Though I'm not sure if I would trust all types to not contain a `"` either.

Tinche:
Yeah, the play might be to just escape any single quotes in the repr.

Tinche:
@phiresky this should be fixed on `main`, can you verify before I release the next version?

phiresky:
Thanks for fixing it! Sorry, it might take a while before I get back to the code that for me generates this error, so probably you shouldn't wait for me.

Tinche:
Confirmed. Will test on 3.7 too before removing the comment.

Tinche:
I fixed it by removing the comment, seems to work fine. Give it one more test for me? <3

[...]

PR #291, opened 2022-08-02
Fix: GenConverter syntax errors with certain types

xaviergmail:
Certain uses of `typing.Literal` or `typing.Annotated` cause syntax errors in generated code. (Related: https://github.com/python-attrs/cattrs/issues/277)

In the case of annotated types (from 3.7 typing_extensions), I get the following syntax error:
```
    eval(compile(script, "", "exec"), globs)
E     File "<string>", line 17
E       raise IterableValidationError('While structuring typing.Dict[typing_extensions.Annotated[str, <class 'housecats.annotations.sanitization.PIIAnnotation'>], vnxpy.jobs.types.leaf.AssigneeRemoval]', errors, __cattr_mapping_cl)
E                                                                                                                     ^
E   SyntaxError: invalid syntax
```

As mentioned in #277 by @Tinche; we _could_ escape quotes manually but using `repr()` on the string is much easier (it returns a string literal)

I tried to write some minimal tests for this but was having some unrelated issues with Poetry/tox that I don't have bandwidth to figure out at the moment. I can however confirm that these changes work when modifying cattrs 22.1.0 in my project's site_packages.

### DIFF
--- a/src/cattrs/gen.py
+++ b/src/cattrs/gen.py
@@ -340,7 +340,7 @@ def make_dict_structure_fn(
             lines.append(f"{i}except Exception as e:")
             i = f"{i}  "
             lines.append(
-                f"{i}e.__note__ = 'Structuring class {cl.__qualname__} @ attribute {an}'"
+                f"{i}e.__note__ = 'Structuring class ' + {cl.__qualname__!r} + ' @ attribute', {an}"
             )
             lines.append(f"{i}errors.append(e)")
 
@@ -352,7 +352,7 @@ def make_dict_structure_fn(
             ]
 
         post_lines.append(
-            f"  if errors: raise __c_cve('While structuring {cl.__name__}', errors, __cl)"
+            f"  if errors: raise __c_cve('While structuring ' + {cl.__name__!r}, errors, __cl)"
         )
         instantiation_lines = (
             ["  try:"]
@@ -360,7 +360,7 @@ def make_dict_structure_fn(
             + [f"      {line}" for line in invocation_lines]
             + ["    )"]
             + [
-                f"  except Exception as exc: raise __c_cve('While structuring {cl.__name__}', [exc], __cl)"
+                f"  except Exception as exc: raise __c_cve('While structuring ' + {cl.__name__!r}, [exc], __cl)"
             ]
         )
     else:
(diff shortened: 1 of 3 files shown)


src/cattrs/gen.py
```diff
@@ -340,7 +340,7 @@ def make_dict_structure_fn(
             lines.append(f"{i}except Exception as e:")
             i = f"{i}  "
             lines.append(
-                f"{i}e.__note__ = 'Structuring class {cl.__qualname__} @ attribute {an}'"
+                f"{i}e.__note__ = 'Structuring class ' + {cl.__qualname__!r} + ' @ attribute', {an}"
```
xaviergmail, line 343: Technically speaking, the concatenation operators aren't required but it reads better in my opinion. Feel free to change if desired.

Tinche:
Thanks for this, I'm kind of stuck implementing converter.copy() atm so don't be discouraged it it takes me a while to come back to this.

src/cattrs/gen.py
xaviergmail: Turns out I had an error in here- I was originally doing `(' ').join(...)` which I botched when moving to string concatenation instead

src/cattrs/gen.py
```diff
@@ -717,7 +717,7 @@ def make_mapping_structure_fn(
             lines.append("      errors.append(e)")
             lines.append("  if errors:")
             lines.append(
-                f"    raise IterableValidationError('While structuring {cl!r}', errors, __cattr_mapping_cl)"
+                f"    raise IterableValidationError('While structuring ' + {repr(cl)!r}, errors, __cattr_mapping_cl)"
```
xaviergmail, line 720: This might look odd at first glance - the first `repr()` gets a string representation of the class and the second repr (`!r`) turns said string into a string literal

Tinche:
Looks good, can you add a small test or two and a changelog snippet?

Tinche:
Hello, since I want to get this release out and you mentioned you'd prefer not to do tests I wrote one myself.

So I'm merging this in now. Thanks!

xaviergmail:
No worries! I have been busier than usual lately. Thank you for wrapping this up. Your work on this project is appreciated!

PR #303, opened 2022-09-12
fix: use __notes__ instead of __note__

### LINKED ISSUE #301: Python 3.11: it's  __notes__, not __note__
cattrs is using `__note__`, but it's actually `__notes__`, and a tuple of strings, and should be (python version conditionally) added via `.add_note(string)`, perhaps? (depending on the backport, I guess). `exceptiongroup` was updated in 1.0.0rc4, see https://github.com/agronholm/exceptiongroup/blob/main/CHANGES.rst.

Also, the message seems pretty poor on pytest + Python 3.10, is pytest and exceptiongroup clashing?

```pytb
      except Exception as e:
        e.__note__ = 'Structuring class Index @ attribute reply'
        errors.append(e)
>     if errors: raise __c_cve('While structuring Index', errors, __cl)
E     cattrs.errors.ClassValidationError: While structuring Index (1 sub-exception)

<cattrs generated structure scikit_build_core.file_api.model.index.Index>:19: ClassValidationError
```

(There's no more helpful err message below it, just that)

I tried running this outside of pytest, and it's much better, I can see the errors. But not inside pytest. I'm guessing it's the monkey patching warnings in https://pypi.org/project/exceptiongroup/. Maybe something to bring up with pytest? Or exceptiongroup? I think that might be https://github.com/agronholm/exceptiongroup/issues/23.

This is really cool when it works, though! (though the notes are missing, due to this issue, unless I force `exceptiongroup==1.0.0rc3`)

henryiii:
Fix #301 by moving from `__note__` to `__notes__`. We could set a minimum `exceptiongroup` version, though if we do that, I think it will enable dev releases even after a non dev release is made.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -21,6 +21,8 @@ History
 * Expose all error classes in the `cattr.errors` namespace. Note that it is deprecated, just use `cattrs.errors`. (`#252 <https://github.com/python-attrs/cattrs/issues/252>`_)
 * ``cattrs.Converter`` and ``cattrs.BaseConverter`` can now copy themselves using the ``copy`` method.
   (`#284 <https://github.com/python-attrs/cattrs/pull/284>`_)
+* Fix usage of notes for the final version of `PEP 678<https://peps.python.org/pep-0678/>`_, supported since ``exceptiongroup>=1.0.0rc4``.
+  (`#303 <303 <https://github.com/python-attrs/cattrs/pull/303>`_)
 
 22.1.0 (2022-04-03)
 -------------------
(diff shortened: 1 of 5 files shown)


Tinche (approved):

Tinche:
LGTM, could you also add a fragment to the release notes?

Tinche:
Thanks!

layday:
`__notes__` is specified as a list of strings in PEP 678 and `BaseException.add_note` operates on lists.  This PR uses tuples, which are incompatible:

```py
In [1]: exc = Exception()

In [2]: exc.add_note("foo")

In [3]: exc.__notes__ = getattr(exc, "__notes__", ()) + ("foo",)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [3], line 1
----> 1 exc.__notes__ = getattr(exc, "__notes__", ()) + ("foo",)

TypeError: can only concatenate list (not "tuple") to list
```

Tinche:
Good catch, should be an easy fix.

henryiii:
Ahh, yes should be easy. Want me to do it?

Tinche:
I think I got it

henryiii:
If it gets added to exceptiongroup, this would be useful: https://github.com/agronholm/exceptiongroup/issues/31

Issue #202, opened 2021-12-26
Insufficient versioned constraint on poetry


sandrotosi:
* cattrs version: 1.9.0
* Python version: 3.10.1
* Operating System: Debian unstable.

### Description

cattrs depends a dependency on poetry as `"poetry-core>=1.0.0"` but that's not strict enough. cattrs uses dependency groups [here](https://github.com/python-attrs/cattrs/blob/main/pyproject.toml#L29) but those are only available in [poetry preview 1.2.0a2](https://github.com/python-poetry/poetry/issues/4938)

please update the versioned depends or revert to a simpler pyproject.toml such as `tool.poetry.dev-dependencies`

Tinche:
Sure, I can bump it up. Do you know to what, though?

sandrotosi:
poetry upstream said '1.2.0a2', but please do note that's a preview release, meaning it has not being released for GA. it may be safer, for the time being, to revert to the previous way to define dev dependencies (so a `tool.poetry.dev-dependencies` table)

Tinche:
That's the version of `poetry`, not `poetry-core` (https://pypi.org/project/poetry-core/#history). Don't see 1.2.a2 for `poetry-core`.

sandrotosi:
you can probably figure that out here: https://github.com/python-poetry/poetry-core/tags / https://pypi.org/project/poetry-core/#history

but once again, these are preview releases, not meant for general usage, ie no distribution/user should have them installed by default only to satisfy the dev-dependencies of cattrs.

Tinche:
Wow, that's a very entitled attitude you got there. I don't care for it one bit.

PR #304, opened 2022-09-13
Fix imprecise dependency on poetry-core

f3flight:
Related to #202 and https://github.com/python-poetry/poetry/issues/4938 - looks like poetry 1.2.0 depends on poetry-core 1.1.0, so I assume the 'groups' feature is present in poetry-core 1.1.0 and that version would be a good one to use.


Tinche:
Thanks!

Issue #308, opened 2022-09-29
cattrs/preconf/json.py is missing import of cattr.preconf.json.JsonConverter


kkg-else42:
* cattrs version: 22.1.0
* Python version: 3.10.6
* Operating System: Win 10

### Description
What I want to do:
```
from cattrs.preconf import json
json_converter: json.JsonConverter = json.make_converter()
```
This doesn't work, because JsonConverter is not imported in cattrs/preconf/json.py

Sure, there are some workarounds, e.g.:
```
from cattr.preconf import json
json_converter: json.JsonConverter = json.make_converter()
```
or
```
from cattrs.preconf import json
from cattrs.preconf.json import JsonConverter
json_converter: JsonConverter = json.make_converter()
```
But I would prefer to be consistent with the cattr**s** namespace.

Tinche:
You're right. Could you put together a quick PR? Maybe also include the others.

kkg-else42:
Gladly, even though I'm new to the project and also new to GitHub.
But I am a little confused, because:
On my machine the installed cattrs-lib shows the working code in `cattr/preconf/json.py` and the imports in `cattrs/preconf/json.py`
On GitHub it is the other way round.

Btw. what exactly do you mean by 
> Maybe also include the others.

Tinche:
It probably got fixed in the mean time (since the last release). You can try installing directly from git and testing it out.

As for the others, I meant the other preconfigured converters (we have a few).

PR #309, opened 2022-09-29
Fix missing imports of preconf converters

kkg-else42:
This fixes missing imports of preconf converters (see: #308)


Tinche:
Nice, thanks!

PR #337, opened 2023-03-10
Add optional dependencies for preconf libraries

alexomics:
This commit adds optional dependencies for the preconf libraries, allowing easier install for downstream users. E.G.:

```console
pip install cattrs[tomlkit]
```

and (I hope) pyproject.toml:

```TOML
[project]
name = "some-package"
version = "1"
dependencies = [
    "cattrs[ujson,tomlkit]"
]
```

It's a small change that moves the preconf dependencies out of the dev group and marks them as optional/extra. It adds an optional dependency on each library and a `preconf` option that installs all extras. 

The change to the tox.ini is needed as I couldn't work out how, with poetry, to add the preconf dependencies back in so that they weren't duplicated. 

I'm sure that I will need to update some docs for this, but I'm not sure where would be best!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,9 +25,25 @@ python = ">= 3.7"
 attrs = ">= 20"
 typing_extensions = { version = "*", python = "< 3.8" }
 exceptiongroup = { version = "*", python = "< 3.11" }
+ujson = { version = "^5.4.0", optional = true }
+orjson = { version = "^3.5.2", markers = "implementation_name == 'cpython'", optional = true }
+msgpack = { version = "^1.0.2", optional = true }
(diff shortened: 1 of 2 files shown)


Tinche:
Very cool, I'm in favor. Although I think the `preconf` extra is overkill and will be an attractive nuisance, bloating up people's dep lists.

Let's add a change log entry for sure, and to the preconf docs page.

alexomics:
Thanks for the quick reply! I've added some explanation to the preconf docs and a changelog entry. Let me know if you think it needs more detail

Tinche:
Looks like we have test failures, let's get that fixed and merge this in! Good job!

Tinche:
Sweet, thanks!

PR #344, opened 2023-03-29
Add reference to Strategies

MicaelJarniac:
This might not be the best way to reference it, but I do believe that there should be a reference to Strategies somewhere in this section.


Tinche:
You're right, that's a good place. But strategies aren't just for unions, so I would link to the tagged union strategy directly instead of strategies in general.

MicaelJarniac:
> You're right, that's a good place. But strategies aren't just for unions, so I would link to the tagged union strategy directly instead of strategies in general.

I've replaced the reference as requested.

Tinche:
Thanks!

Issue #347, opened 2023-04-06
Support both Attribute.name and AttributeOverride.rename 


getglad:
### TLDR 

I'd like to have a `rename` ergonomic such as `override(rename="<FieldName>", accept_both_names=True)` that supports both the actual attribute as well as the `rename` mapping. I have a working implementation that I could try to convert into a PR if this strategy fits.

### Description

1. I have a scenario where I will have a set of input key names that will be changed, and for some period of time I expect to have have both formats in my system. I will still need to be able to `structure` both of them into an `attr`.
2. I would like to be able to convert the `attr` into the new format and make all of my internal changes without doing much `post_init` work in the `attr` (ie, populating an old field into a new field if old field was used).

In a simple example, my format will convert from `{'OldField':  'value1'}` to `{'NewField': 'value2'}`.

And my `attr` would change from

```
@define
class MyInternalAttr:
    OldField: str
```
to 

```
@define
class MyInternalAttr:
    NewField: str
```

If I only needed to cast `OldField` -> `NewField`, I know I could use `make_dict_structure_fn` with  `NewField=override(rename="OldField")`, but at present I could not structure both `OldField` and `NewField`, at least not on the same converter.

My goal is along the lines of what [rename](https://catt.rs/en/stable/customizing.html#rename) can already support, except that rename operates as a full override - I can have the old field or the new field but not both.

### Suggested Implementation

What I'd like to have is an ergonomic along the lines of `NewField=override(rename="OldField", accept_both_names=True)`

I have some stubbed out changes that seem to work, but wanted to check in before I worked towards a PR. The below is a summary of changes I had to make in `cattrs.gen`.

#### Update `AttributeOverride` and the `override` method

```
@frozen
class AttributeOverride:
    omit_if_default: Optional[bool] = None
    rename: Optional[str] = None
    omit: bool = False  # Omit the field completely.
    accept_both_names: bool = False


def override(
    omit_if_default: Optional[bool] = None,
    rename: Optional[str] = None,
    omit: bool = False,
    accept_both: bool = False,
):
    return AttributeOverride(
        omit_if_default=omit_if_default,
        rename=rename,
        omit=omit,
        accept_both_names=accept_both,
    )
```

##### Example in Current Release

https://github.com/python-attrs/cattrs/blob/405f0291b958ae9eb45ee38febeb91fb65dd644f/src/cattrs/gen.py#L336-L350

##### Code Change

[...]

PR #351, opened 2023-04-11
fallback key name usage example docs

getglad:
Providing examples as discussed in https://github.com/python-attrs/cattrs/issues/347

Note: I can tell there is some formatting applied to the existing docs, but I couldn't find guidance in `CONTRIBUTING.md` or a clear command in `Makefile` - but happy to take some pointers and add those too.

Hope this is along the lines we discussed!


Tinche (approved):
LGTM, I might reword it a little after merging.

Tinche:
Thanks!

PR #353, opened 2023-04-12
Allow overriding unstruct_collection_overrides

### LINKED ISSUE #350: Allow overriding unstruct_collection_overrides in preconf make_converter
* cattrs version: 22.2.0
* Python version: 3.9.16
* Operating System: macOS

### Description

What I was trying to do:
```python
from collections.abc import Set

from cattrs.preconf.json import make_converter as json_make_converter

json_converter = json_make_converter(unstruct_collection_overrides={Set: sorted})
```

This didn't work because `cattrs.preconf.json.make_converter` overrides the `kwargs`: https://github.com/python-attrs/cattrs/blob/5ccd616cb9ae2ad88b5d4ba7816f7fad3a9f84be/src/cattrs/preconf/json.py#L40-L44

Then I tried:
```python
from collections.abc import Set

from cattrs.preconf.json import make_converter as json_make_converter

json_converter = json_make_converter().copy(unstruct_collection_overrides={Set: sorted})
```

But later in the code I call `json_converter.dumps` and mypy complained `error: "Converter" has no attribute "dumps"  [attr-defined]` because `copy` returns a `Converter`.

### What I Did
```python
from collections.abc import Set

from cattrs.preconf.json import JsonConverter
from cattrs.preconf.json import configure_converter as json_configure_converter

json_converter = JsonConverter(unstruct_collection_overrides={Set: sorted})
json_configure_converter(json_converter)
```

Maybe the two forms that I've tried above should also be allowed?

ericbn:
Fixes #350

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -14,6 +14,8 @@
   ([#81](https://github.com/python-attrs/cattrs/issues/81))
 - Add `cbor2` serialization library to the `cattr.preconf` package.
 - Add optional dependencies for `cattrs.preconf` third-party libraries. ([#337](https://github.com/python-attrs/cattrs/pull/337))
+- All preconf converters now allow overriding the default `unstruct_collection_overrides` in `make_converter`.
+  ([#350](https://github.com/python-attrs/cattrs/issues/350) [#353](https://github.com/python-attrs/cattrs/pull/353))
 
 ## 22.2.0 (2022-10-03)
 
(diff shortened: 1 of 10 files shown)


Tinche:
Lgtm, just add a small changelog snippet and we're good to go. Thanks!

ericbn:
Cool, done!

Tinche:
Thanks!

PR #365, opened 2023-05-22
ISO-8601 support for preconfigured converters

GustasG:
Serialize and deserialize dates and times to ISO-8601 format in preconfigured converters. Only added this functionality to those preconfigured converters that already use datatime conversion to/from ISO-8601


Tinche:
Howdy,

sorry for the delay. Sure, let's get this in, but it'll need a few tests first. And please rebase on `main` first.

Tinche:
This was superceeded by https://github.com/python-attrs/cattrs/pull/420, but thank you for getting it up to here!

PR #366, opened 2023-05-22
Allows use of `typing_extensions.Annotated` when python version is less than `3.9`

jdoiro3:
Currently, the use of `typing_extensions.Annotated` doesn't work. This PR adds support for the type when the version of Python is less that `3.9` (i.e., the user can't use `typing.Annotated`).

### DIFF
--- a/src/cattrs/_compat.py
+++ b/src/cattrs/_compat.py
@@ -175,6 +175,7 @@ if is_py37 or is_py38:
     from collections import Counter as ColCounter
     from typing import Counter, Union, _GenericAlias
     from typing_extensions import Annotated, NotRequired, Required
+    from typing_extensions import get_origin as te_get_origin
 
     if is_py38:
         from typing import TypedDict, _TypedDictMeta
@@ -183,7 +184,7 @@ if is_py37 or is_py38:
         TypedDict = ExtensionsTypedDict
 
     def is_annotated(type) -> bool:
-        return get_origin(type) is Annotated
+        return te_get_origin(type) is Annotated
 
     def is_tuple(type):
         return type in (Tuple, tuple) or (


Tinche:
Cool! Can you add a changelog entry too? And maybe a line to the Annotated sections in the docs?

jdoiro3:
> Cool! Can you add a changelog entry too? And maybe a line to the Annotated sections in the docs?

Just added an entry under the current unreleased version. Let me know if any other changes are needed.

src/cattrs/_compat.py
```diff
@@ -143,9 +143,10 @@ def get_final_base(type) -> Optional[type]:
 
     from collections import Counter as ColCounter
     from typing import Counter, Union, _GenericAlias
+    from typing_extensions import Annotated, get_origin
 
-    def is_annotated(_):
-        return False
+    def is_annotated(type) -> bool:
+        return get_origin(type) == Annotated
```
Tinche, line 149: I think you can use `is` here?

Tinche (changes requested):
Please rebase and let's get this merged in.

src/cattrs/_compat.py
jdoiro3: Fixed.

jdoiro3:
@Tinche, not sure why CI isn't running but this looks good to go.

Tinche:
I reran the action, looks like there's a linting issue now.

jdoiro3:
@Tinche, can you run CI again? I removed the `get_origin` import from `typing_extensions` since it seems to be the same implementation as `typing`. Wasn't sure it was.

jdoiro3:
Looks like everything passes except `pre-commit.ci`.

Tinche:
Sweet. I don't actually know where the pre-commit.ci is defined. Maybe the python-attrs organization?

Tinche (approved):

PR #371, opened 2023-05-31
PDM switch

Tinche:


JWCook:
@Tinche Just curious, what motivated you to switch from poetry to PDM?

I've also been looking into other packaging tools recently, and what I'd really like is something like poetry + PEP 621 support, or hatch + lockfile support. PDM looks appealing, and I'd be interested to hear what you think of it so far.

Tinche:
I find that PDM has better defaults (poetry's `~` operator imposes an upper bound which I think is to be discouraged for library dependencies), follows standards more closely and is better maintained.

Hatch + a lockfile would be a strong contender though. I also enjoy that PDM has development dependencies that aren't extras, so they don't get published to PyPI. Also a cool feature is that PDM can be used with the hatchling backend, so hatch(ling) plugins can be used (as cattrs does). I think that's neat and composable.

JWCook:
Ah, I wasn't aware it could use other build backends and their plugins. Cool! Thanks for the info.

PR #378, opened 2023-06-07
Fix type checker error

layday:
Pyright expects `typing.Type` to be parameterised and reports that the function's type is unknown.


Tinche:
LGTM, let's just add a small changelog snippet.

Tinche:
Thanks!

PR #379, opened 2023-06-08
Make `BaseValidationError` base static in Python < 3.11

layday:
Unsure if this worked in MyPy but Pyright failed to resolve the base of `BaseValidationError` subclasses leading to hard to trace type errors in custom note attributes from failing to narrow Iterable and ClassValidationError unions.


Tinche:
Cool, thanks!

PR #382, opened 2023-06-11
Make Python versions known statically

layday:
Type checkers are unable to interpret `sys.version_info` aliases and generate unions between Python version branches.  If either branch depends on packages which are not installed the union might be reduced to an unknown or any type.

Follow-up from https://github.com/python-attrs/cattrs/pull/379 (Pyright would merge the two branches).

### DIFF
--- a/src/cattrs/_compat.py
+++ b/src/cattrs/_compat.py
@@ -30,25 +30,10 @@ try:
 except ImportError:
     ExtensionsTypedDictMeta = None
 
-__all__ = [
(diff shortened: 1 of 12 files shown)


Tinche:
I'm willing to take this, but what's the benefit? I don't type check the insides of cattrs since it's very meta-programmy.

Will users of Pyright have a better time with these changes?

Tinche:
Thanks!

PR #385, opened 2023-06-13
Don't try to upload coverage if CI is run via workflow dispatch on a fork

AlexWaygood:
While investigating https://github.com/python/typing_extensions/issues/230 and preparing #384, I tried running cattrs's CI on my fork via workflow-dispatch. This was made a little more complicated by the fact that the "upload coverage to codecov" step always seems to fail when the workflow is run on a fork


Tinche:
Thanks! I want to get rid of codecov long-term but still looking for a good alternative.

PR #386, opened 2023-06-14
Don't cancel all jobs in CI if one job fails

AlexWaygood:
This will make it easier to see in #384 which tests (if any) are actually failing on which Python versions -- currently the codecov failures are just leading to all jobs being cancelled in CI


Tinche:
Thanks!

PR #396, opened 2023-07-19
docs: add strategy using dedicated (un)structure method

pohlt:
Feel free to edit. 👍🏼

### DIFF
--- a/docs/strategies.md
+++ b/docs/strategies.md
@@ -258,3 +258,34 @@ Child(a=1, b='foo')
 ```{versionadded} 23.1.0
 
 ```
+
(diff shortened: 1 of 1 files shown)


Tinche:
Howdy,

good job. We've had a small misunderstanding though; strategies are basically functions you can import and call, and they apply a potentially complex customization on your behalf. What you have here is a piece of documentation that looks like a recipe.

That's also good! It should just go elsewhere. Can you move it to the common usage examples page instead?

pohlt:
Ok, got it. Will take some time, but I'll do as requested.

pohlt:
Is included in #405, so closing here.

PR #400, opened 2023-07-28
Broaden method type definition for preconf orjson converter.

PIG208:
`bytes` is not the only input type `orjson.loads` allows.


Tinche:
Sweet, thanks!

PR #404, opened 2023-08-03
Fix typos for notes getattr calls

PIG208:
The intended first argument to `getattr` should be the exception in question.


Tinche:
You're right. Should probably test for these too 😬

PR #406, opened 2023-08-04
Implement NoneType structuring.

PIG208:
This is useful to have for custom structure hooks of union types that are also optional.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -26,6 +26,8 @@
   ([#398](https://github.com/python-attrs/cattrs/issues/398) [#399](https://github.com/python-attrs/cattrs/pull/399))
 - Broaden loads' type definition for the preconf orjson converter.
   ([#400](https://github.com/python-attrs/cattrs/pull/400))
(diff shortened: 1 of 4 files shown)


Tinche:
I've been asked for this before and have been reluctant to add it since I don't want cattrs to become a kitchen sink and I've personally not needed it in the last 10 years (and I use cattrs a lot). Can you elaborate on the usecases you think it'd be helpful in?

PIG208:
I was using it when recursively calling `converter.structure` on the type args of `Union`. With this structure function I can do something like this:
```python
def structure_union(value: object, target_type: Type[T]) -> T:
    type_args = get_args(targe t_type)
    errors = []
    for type_arg in type_args:
        try:
            return json_converter.structure(value, type_arg)
        except Exception as exc:
            errors.append(exc)
    raise UnionError(f"{value} cannot be structure as {target_type}", errors, target_type)
```
In hindsight, I think it is also quite easy to just special case `NoneType` or register a custom hook of my own. So it is not strictly necessary.

PR #408, opened 2023-08-05
Make validation notes picklable.

PIG208:
Certain test runners (in my case, Django's `RemoteTestRunner`) pickle the exceptions raised to accumulate them from multiple processes.

For exceptions with `AttributeValidationNote` and/or `IterableValidationNote` attached, because we have overridden `__new__` on these classes, pickle cannot correctly unpickle them with the required arguments. For this use case, exceptions being pickable is necessary to accumulate errors with the `multiprocessing` setup.

This implements [`__getnewargs__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs__) to fix this problem.


Tinche:
Thanks!

PR #415, opened 2023-08-12
Fix typeddict tests on systems with 32-bit time_t

mgorny:
Reduce the range of generated `datetime` instances to values valid for 32-bit `time_t` range, as otherwise multiple tests fail with errors such as:

    FAILED tests/test_typeddicts.py::test_simple_roundtrip - OverflowError: timestamp out of range for platform time_t

This is based on an earlier fix for `test_preconf.py`, see a0e56f43f061c43814d6f938833d1c325ed61525
and c58028789454fc7a9b459b94c214cab2ab1acb81.

Originally reported as https://bugs.gentoo.org/912187.


Tinche:
Cool, thanks!

mgorny:
Thanks!

PR #416, opened 2023-08-14
Use the un-underscored name if no alias attribute is there.

jap:
This appears to be caused by a thinko in #391, this fixes it again.

### DIFF
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -17,7 +17,7 @@ from attrs import Factory, define, fields, make_class
 from hypothesis import HealthCheck, assume, given, settings
 from hypothesis.strategies import booleans, just, lists, one_of, sampled_from
 
-from cattrs import Converter, UnstructureStrategy
+from cattrs import BaseConverter, Converter, UnstructureStrategy
 from cattrs.errors import ClassValidationError, ForbiddenExtraKeysError
 from cattrs.gen import make_dict_structure_fn, override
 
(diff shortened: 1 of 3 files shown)


Tinche:
Ah, cool. Can you add a changelog note? And a simple test would be great so it doesn't happen again.

jap:
Actually -- given that `cattrs` depends on modern `attrs` and this is just compat code, let me simplify it!

Tinche:
Howdy, any progress on this? Trying to plan the next release.

jap:
Whoops, sorry - forgot about this PR, working on an updated version now!

tests/test_converter.py
```diff
@@ -381,6 +381,23 @@ class C:
     assert inst == converter.structure(unstructured, C)
 
 
+def test_dict_roundtrip_with_alias():
+    """
+    A class with an aliased attribute can be unstructured and structured.
+    """
+
+    converter = Converter()
```
Tinche, line 389: I think you should be testing this with a `BaseConverter` to hit the changed code path, the `Converter` uses different machinery.

Tinche (changes requested):
Looks good, left a comment

Tinche:
Just a small note, I'll be away for the next 3 weeks on a family trip and I can continue reviewing this after I'm back!

Tinche:
I want to get the next release out so I fixed this myself. Thanks!

PR #420, opened 2023-08-21
Add date to preconfigured converters

Peter554:
Adding support for `date` in the preconfigured converters. I see there was already some effort made in this direction here https://github.com/python-attrs/cattrs/pull/365, however that PR is missing tests and seems to be a bit inactive.


src/cattrs/preconf/msgpack.py
```diff
@@ -30,6 +30,8 @@ def configure_converter(converter: BaseConverter):
     converter.register_structure_hook(
         datetime, lambda v, _: datetime.fromtimestamp(v, timezone.utc)
     )
+    converter.register_unstructure_hook(date, lambda v: v.isoformat())
```
Peter554, line 33: I don't know much about msgpack, but it seems like the msgpack converter is doing things a bit differently with `datetime` - it is using `timestamp` rather than `isoformat`. Is there a reason the msgpack converter does this differently? How should we be handling `date`s in the msgpack converter?

Peter554:
Not sure what I have to do to get CI to run, but `make lint` and `make test` are passing for me locally.

src/cattrs/preconf/msgpack.py
Tinche: We can, but let's make it use timestamps too, for consistency.

The idea is for the JSON converters to use the ISO format since JSON is meant to be somewhat human readable, and JavaScript can parse them out of the box. For binary formats like msgpack I think timestamps are a better fit since they're much more efficient.

Tinche:
Thanks for picking this up. I've clicked the button to run the tests, it needs my approval I guess.

tests/test_preconf.py
```diff
@@ -148,6 +150,7 @@ def everythings(
         Everything.AnIntEnum.A,
         Everything.AStringEnum.A,
         draw(dts),
+        draw(dates()),
```
Tinche, line 153: Please set the max value here similar to what the datetime strategy is doing above. If we don't do this, we might break some 32-bit Linux downstreams (don't ask...).

Peter554: Ah okay 😬  done

src/cattrs/preconf/msgpack.py
Peter554: Interesting, thanks. Updated now. Since `date` has no `.timestamp` method, I've used the midnight-aligned datetime.

Tinche:
Great job, thanks!

PR #431, opened 2023-09-23
Fix include_subclass union_strategy typing

VincentVanlaer:
Using Converter instead of BaseConverter as one of the arguments to union_strategy allows more strategies to be passed. Specifically, the configure_tagged_union strategy currently requires a Converter according to its typing.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -51,6 +51,8 @@ Our backwards-compatibility policy can be found [here](https://github.com/python
 - The documentation has been significantly reworked.
   ([#473](https://github.com/python-attrs/cattrs/pull/473))
 - The docs now use the Inter font.
+- Make type annotations for `include_subclasses` and `tagged_union` strategies more lenient.
+  ([#431](https://github.com/python-attrs/cattrs/pull/431))
 
 ## 23.2.3 (2023-11-30)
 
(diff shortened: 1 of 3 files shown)


Tinche:
Hm how about we change the tagged union strategy to take a BaseConverter instead? I don't remember why I made it take a Converter, maybe it's an easy fix.

VincentVanlaer:
It seems like the tagged union strategy only uses methods on `BaseConverter`, so changing that seems fine. In the meantime I realized it might make more sense to type `include_subclasses` as

```python
C = TypeVar("C", bound=BaseConverter)

def include_subclasses(
    cl: Type,
    converter: C,
    subclasses: Optional[Tuple[Type, ...]] = None,
    union_strategy: Optional[Callable[[Any, C], Any]] = None,
    overrides: Optional[Dict[str, AttributeOverride]] = None,
) -> None:
```
since the argument passed to the union strategy is the same as is passed to `include_subclasses`. But that might be overkill for this situation. Let me know whether you prefer just the change to the tagged union strategy or also for `include_subclasses`.

Tinche:
I'd be ok with changing both!

Sorry for the delay, I was focused on getting 23.2 out the door which is now done.

VincentVanlaer:
No worries! I have included both changes and I checked with mypy that no new errors emerged.

Tinche:
Sweet! Can you add a small line to HISTORY?

VincentVanlaer:
I missed your last comment! Should be ready now.

Tinche:
Thanks!

Issue #426, opened 2023-09-17
include_subclasses does not support dataclasses


eltoder:
* cattrs version: 23.1.2
* Python version: 3.9
* Operating System: Linux

### Description

According to [the documentation](https://catt.rs/en/stable/cattrs.strategies.html#cattrs.strategies.include_subclasses), include_subclasses() supports both attrs and dataclasses. However, when I change [the example](https://catt.rs/en/stable/strategies.html#include-subclasses-strategy) to use dataclasses, I get an error:

```python
from dataclasses import dataclass
from cattrs.strategies import include_subclasses
from cattrs import Converter

@dataclass
class Parent:
    a: int

@dataclass
class Child(Parent):
    b: str

converter = Converter()
include_subclasses(Parent, converter)
```
produces
```
Traceback (most recent call last):
  File "/home/eltoder/dev/scratch/cattrs_dc.py", line 14, in <module>
    include_subclasses(Parent, converter)
  File "/home/eltoder/.local/share/ext-python/python-3.9.16.146/lib/python3.9/site-packages/cattrs/strategies/_subclasses.py", line 73, in include_subclasses
    _include_subclasses_without_union_strategy(
  File "/home/eltoder/.local/share/ext-python/python-3.9.16.146/lib/python3.9/site-packages/cattrs/strategies/_subclasses.py", line 111, in _include_subclasses_without_union_strategy
    dis_fn = converter._get_dis_func(subclass_union)
  File "/home/eltoder/.local/share/ext-python/python-3.9.16.146/lib/python3.9/site-packages/cattrs/converters.py", line 756, in _get_dis_func
    return create_uniq_field_dis_func(*union_types)
  File "/home/eltoder/.local/share/ext-python/python-3.9.16.146/lib/python3.9/site-packages/cattrs/disambiguators.py", line 20, in create_uniq_field_dis_func
    cls_and_attrs = [
  File "/home/eltoder/.local/share/ext-python/python-3.9.16.146/lib/python3.9/site-packages/cattrs/disambiguators.py", line 21, in <listcomp>
    (cl, set(at.name for at in fields(get_origin(cl) or cl))) for cl in classes
  File "/home/eltoder/.local/share/ext-python/python-3.9.16.146/lib/python3.9/site-packages/attr/_make.py", line 1944, in fields
    raise NotAnAttrsClassError(f"{cls!r} is not an attrs-decorated class.")
[...]

PR #444, opened 2023-11-15
Add a regression test for a nested tagged union scenario

bluetech:
I've converted some project which uses cattrs from attrs to dataclasses, but that failed because the union dismbiguator doesn't support dataclasses yet (https://github.com/python-attrs/cattrs/issues/426#issuecomment-1731787790). So I used the opportunity to convert the code to the new tagged union strategy which is much nicer than the disambiguation stuff anyway. One case looks like the test added in this PR, which failed too (cattrs  v23.1.2). Looking into it I saw that the problem has been fixed in main by commit b25a25800f245d4504f8d9aaaa90faeca3e4adbb. So I'm awaiting a new release, but in the meantime thought this regression test might be helpful.

### DIFF
--- a/tests/strategies/test_tagged_unions.py
+++ b/tests/strategies/test_tagged_unions.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Optional, Sequence, Union
 
 from attrs import define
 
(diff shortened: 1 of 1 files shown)


Tinche:
Cool, thanks. The next release is imminent!

PR #459, opened 2023-11-22
Guess we doin V now

Tinche:


Tinche:
@petergaultney this is very much experimental and a work-in-progress at this point ;)

MicaelJarniac:
Is there an issue, or anything, describing what's going on?

MicaelJarniac:
From a quick glance, this seems to be the new feature:
```py
    hook = customize(
        c,
        Small,
        V((fs := f(Small)).a).ensure(greater_than(10)),
        V(fs.b).ensure(len_between(0, 10)),
    )
```
I'm pretty clueless as to what's going on, but may I suggest also supporting something like:
```py
    hook = customize(
        c,
        Small,
        V((fs := f(Small)).a) > 10 and
        0 <= len(V(fs.b)) <= 10,
    )
```
By overwriting the `__gt__`, `__len__`, `__and__`, and such operators, to construct these validations in a more fluent way?

This would be similar to how libs like [SQLAlchemy handle some queries](https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html#).

These operators could return the same that `.ensure(greater_than(10))` would return, for example, and not `bool`.

Issue #445, opened 2023-11-16
High CPU usage after upgrade


sahu-sunil:
* cattrs version: 23.1.2
* Python version: 3.9
* Operating System: Linux

We are using cattrs for config validation purpose of our tests in Asynchronous framework. Thousands of tests run every minute. We upgraded from 1.5.0 to 23.1.2 which has major changes on how cattrs back-end works. The cpu touches 100% in couple of mins. After profiling we found that cache (linecache) was the reason for high cpu usage. We disabled the caching (including detailed validations) with `_cattrs_use_linecache=False` which helped a bit, now it takes couple of hours (3-4h) to reach cpu to 100.
We see `generate_unique_filename` is still taking lot of time.
![image](https://github.com/python-attrs/cattrs/assets/18591767/537525b9-3cbf-4b16-9c17-abc0bde89ade)

Is there any other way to handle this in cattrs latest version?

Tinche:
Interesting. That's quite a large upgrade though, a ton of things have changed. Are you using `cattrs.Converter`, `cattrs.BaseConverter` or functions imported directly from cattrs (like `cattrs.structure`)?

You can likely resolve your issue by switching to `cattrs.BaseConverter`, that converter class does minimal code generation. It's somewhat slower to structure and unstructure though. Would be curious for more information!

sahu-sunil:
I am using `cattrs.Converter` and then registering hooks like -
```
converter = Converter(detailed_validation=False)
converter.register_structure_hook(
    ClassName,
    make_dict_structure_fn(
    ClassName,
    converter,
    _cattrs_use_linecache=False,
    _cattrs_detailed_validation=False,
    ),
)
converter.structure(....)
```

Will try `BaseConverter` for less initialisation but AFAIK it will still go for functions mentioned in above profiler snapshot.

Tinche:
BaseConverter won't help if you're using `make_dict_structure_fn`directly, since that's the source of the issue.

Well, the root issue is that you're probably recreating the same class a lot in your tests.

Let me give you some context. `make_dict_structure_fn` will compile a specialized function for structuring. It'll also store the source code of that function using Python's [linecache module](https://docs.python.org/3/library/linecache.html); this is necessary for good stack traces and for stepping through the function in a debugger.

[...]

PR #461, opened 2023-11-23
Avoid linecache unique file generation in case of linecache disabled

sahu-sunil:
When cattrs' converter/hooks or same classes are created too frequent or in large amount, `generate_unique_filename` (generation of unique file with uuid) becomes slow/inefficient even if `linecache` flag is disabled.
This PR executes this function only when cache is enabled. More optimisation can be done for compilation of same classes repeatably, as mentioned in below issue.
Issue discussion: https://github.com/python-attrs/cattrs/issues/445

### DIFF
--- a/src/cattrs/gen/__init__.py
+++ b/src/cattrs/gen/__init__.py
@@ -212,22 +212,20 @@ def make_dict_unstructure_fn(
             + ["  return res"]
         )
         script = "\n".join(total_lines)
-
-        fname = generate_unique_filename(
-            cl, "unstructure", reserve=_cattrs_use_linecache
-        )
-
-        eval(compile(script, fname, "exec"), globs)
-
-        fn = globs[fn_name]
+        fname = ""
         if _cattrs_use_linecache:
+            fname = generate_unique_filename(
+                cl, "unstructure", reserve=_cattrs_use_linecache
+            )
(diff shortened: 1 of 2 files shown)


Tinche:
@sahu-sunil cool, I can merge this in. Does it fix your actual problem though? You can install from this branch and test.

If not I can try applying more optimizations.

sahu-sunil:
Yes it was bit helpful but in long duration CPU is still gradually increasing. I am confident that this is due to `linecache` left enabled somewhere hidden place. I am going to try `cattrs` with linecache `False` by default instead of `True` (current behaviour). 

Yes, please check if you see some scope of optimisation on repeated compilation

Tinche:
@sahu-sunil Ok, let's merge this in and I can take a look at a bigger optimization.

Can you just add a short line to HISTORY.md?

Tinche:
Thanks, I'll ping you when I have something more.

sahu-sunil:
> Yes it was bit helpful but in long duration CPU is still gradually increasing. I am confident that this is due to `linecache` left enabled somewhere hidden place. I am going to try `cattrs` with linecache `False` by default instead of `True` (current behaviour).

I tried with this locally and the cpu is stable now. Will see if I can find all the places where linecache can be disabled.

Issue #489, opened 2024-01-16
Constructor selection for structuring


AdrianSosic:
* cattrs version: 23.2.3
* Python version: 3.9
* Operating System: macOS

### Description

I'm not sure if this idea has already come up somewhere, or if there's perhaps a much smarter/different way to approach the problem, but I wanted to ask if there's already a mechanism for specifying alternative constructors for structuring.

Consider the following class:
```python
@define
class Point:
    x: float
    y: float

    @classmethod
    def from_polar(cls, radius: float, angle: float) -> Point:
        return Point(radius * math.cos(angle), radius * math.sin(angle))
```
Let's assume it is used somewhere down the line a nested object hierarchy, and potentially that the class could have many other `classmethod` constructors. I am looking for a way to specify the constructor in the structuring input.

### What I Did

Here is a relatively simple hook that does the job
```python
def select_constructor_hook(specs: dict, cls: Type[_T]) -> _T:
    """Use the constructor specified in the 'constructor' field for deserialization."""
    # If a constructor is specified, use it
    specs = specs.copy()
    if constructor_name := specs.pop("constructor", None):
        constructor = getattr(cls, constructor_name)

        # Extract the constructor parameter types and deserialize the arguments
        type_hints = get_type_hints(constructor)
        for key, value in specs.items():
            annotation = type_hints[key]
            specs[key] = cattrs.structure(specs[key], annotation)

        # Call the constructor with the deserialized arguments
        return constructor(**specs)

    # Otherwise, use the regular __init__ method
    return cattrs.structure_attrs_fromdict(specs, cls)
```
It allows to create an object as follows
```python
cattrs.register_structure_hook(Point, select_constructor_hook)
cattrs.structure({"constructor": "from_polar", "radius": 1.0, "angle": math.pi}, Point)
```

### The Question
To me, this appears like a rather common use case and I am wondering if there is already a similar/better mechanism in place. If not, would this be a pattern worth to be included into cattrs, potentially in the form of a strategy? Of course, the exact details would be need to fleshed out and things should become configurable (e.g. which keyword to use for the constructor specification, etc). 

I could imagine something like
```python
from cattrs.strategies import switch_constructors
switch_constructors("constructor", converter)
```

Happy to hear your thoughts 😃

Tinche:
Have you seen https://catt.rs/en/stable/strategies.html#using-class-specific-structure-and-unstructure-methods? Feels like it matches this use case.

Here's how this would look like using that strategy:
```python
[...]

PR #494, opened 2024-02-01
Recipes for initializer selection

AdrianSosic:
Hi @Tinche,

here's a first draft for the "Recipes" page you suggested in #489.
Let me know what you think and what needs to be adapted. Of course feel free to adjust examples and text as you like.

BTW: Should we add some tests for the recipes, or is there some automatic doc testing already in place (haven't yet looked too much into the repo mechanisms, tbh).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -19,6 +19,7 @@ caption: User Guide
 
 customizing
 strategies
+recipes
 validation
 preconf
 unions
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -1,0 +1,124 @@
+# Recipes
+
+This page contains a collection of recipes for custom un-/structuring mechanisms.
+
+## Switching Initializers
+
+When structuring _attrs_ classes, _cattrs_ uses the classes' ``__init__`` method to
+instantiate objects by default.
+In certain situations, you might want to deviate from this behavior and use
+alternative initializers instead.
+
+For example, consider the following `Point` class describing points in 2D space,
+which offers two `classmethod`s for alternative creation:
+```python
+from __future__ import annotations
+
+import math
+
+from attrs import define
+
+
+@define
+class Point:
+    """A point in 2D space."""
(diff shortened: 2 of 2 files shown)


Tinche:
I like it!

Do you think we should export `StructureHook` (and `UnstructureHook`) from `cattrs` directly?

> Should we add some tests for the recipes

You can use doctests. See an example here: https://github.com/python-attrs/cattrs/blob/3c4572fd28e587c02c1112904c981ff3a6c3801c/docs/customizing.md?plain=1#L68. You can run them using `make doctests`, and they get run on RTD when the docs get built. If they fail they stop the build, making someone take a look.

Also, I've been trying to use semantic newlines for docs, like _attrs_: https://github.com/python-attrs/attrs/blob/main/.github/CONTRIBUTING.md#documentation. In practice, this means lines break not based on line length but a natural break (like a comma).

AdrianSosic:
Hi @Tinche,

you mean making `StructureHook` and `UnstructureHook` available directly from the high-level `cattrs` namespace? I think that sound like a good idea. Would make it easier for people who prefer to annotate their hooks (like me 🙃).

I've tried my best to incorporate the requested changes. Was not 100% sure about the semantic newlines but tried to follow the style from the other `.md` files in the docs folder.

Let me know if you see other things that require changes (e.g. concerning the hook imports mentioned above). Otherwise, this PR is ready from my side 👌🏼

Tinche:
> you mean making StructureHook and UnstructureHook available directly from the high-level cattrs namespace

Yeah, exactly.

In any case, great work. I might do an editing pass before releasing the next version.

Tinche:
Fixed it! Looks like `doctest` blocks don't do syntax highlighting unless they are written in the prompt style :/

AdrianSosic:
Hm, seems inconvenient... but thanks for the fix anyway!! ✌️

PR #507, opened 2024-02-19
Don't install `msgspec` on PyPy

jonathanberthias:
`msgspec` is [only supported on CPython](https://github.com/jcrist/msgspec/issues/22#issuecomment-1349049009). This creates an issue when installing with: `pdm install -G :all`.


Tinche:
Thanks!

PR #508, opened 2024-02-19
Fix typo in strategies.md

EFord36:
While reading the docs, I just noticed this small typo, so thought I would submit a quick PR.


Tinche:
Good catch, thanks!

Issue #513, opened 2024-02-27
Documentation does not explain unstruct_collection_overrides' input


td-anne:
* cattrs version: 23.2.3
* Python version: 3.11.4
* Operating System: MacOS

### Description

I am trying to unstructure sets into something that JSON can cope with *and* that is canonical (equal for any two equal sets; set iteration order can differ for equal sets). I tried `unstruct_collection_overrides={Set: sorted}` and obtained a very confusing error ("< not supported between dict and dict"). It turns out that the problem is that the functions in `unstruct_collection_overrides` receive a generator that emits the unstructured contents of the collection (at least for Set; I'm not sure for dicts). Since the objects in my set are frozen attrs objects, they themselves work fine in a Set (and being sorted), but once unstructured they become dictionaries, which can't be sorted.

The problem: the documentation at https://catt.rs/en/stable/unstructuring.html#customizing-collection-unstructuring never specifies what the unstructuring functions receive as input. Neither does the docstring for Converter (https://catt.rs/en/stable/cattrs.html#cattrs.Converter). A little explanation would have helped tremendously in tracking down the mysterious error.

### What I Did

```python
@frozen
class A:
    x: int = 0
    y: int = 0


def convert_set(s):
    list_s = list(s)
    print(f"{list_s=}")
    return sorted(list_s)


converter = Converter(unstruct_collection_overrides={Set: convert_set})

converter.unstructure({A(x=1), A(x=2), A(x=3)})
```

gives

```
list_s=[{'x': 3, 'y': 0}, {'x': 1, 'y': 0}, {'x': 2, 'y': 0}]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], [line 15](vscode-notebook-cell:?execution_count=12&line=15)
     [10](vscode-notebook-cell:?execution_count=12&line=10)     return sorted(list_s)
     [13](vscode-notebook-cell:?execution_count=12&line=13) converter = Converter(unstruct_collection_overrides={Set: convert_set})
---> [15](vscode-notebook-cell:?execution_count=12&line=15) converter.unstructure({A(x=1), A(x=2), A(x=3)})

[...]

PR #515, opened 2024-03-04
WIP: Expand docs on unstructuring

td-anne:
I ran into some confusing situations trying to unstructure some objects I had defined (#513 ) and it was suggested I create a PR clarifying the docs. This PR attempts to do this.

I should say, my understanding of how `cattrs` works, and of how `cattrs` is meant to be used, is a little shaky, so feedback on this documentation is highly welcome.

I am also totally new to `cattrs` development, so I'm not sure where anything goes. In particular, is there a place for demo code that will be executed? Should I add tests to verify that the behaviour I am describing in the docs is really what happens? I'm not sure how to run the doctests, particularly in the `docs/` directory.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -72,7 +72,7 @@ This new hook can be used directly or registered to a converter (the original in
 ```
 
 
-Now if we use this hook to structure a `Model`, through ✨the magic of function composition✨ that hook will use our old `int_hook`.
+Now if we use this hook to structure a `Model`, through ✨the magic of function composition✨ that hook will use the pre-existing (default) `Model` hook, which takes advanatge of our `int_hook`.
 
 ```python
 >>> converter.structure({"a": "1"}, Model)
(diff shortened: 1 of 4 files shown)


Tinche:
Hi!

I completely dropped the ball on this and I apologize for that, I got sidetracked by other things and forgot about this.

I will try to integrate this into cattrs soon as the last thing I do before releasing the next version.

> In particular, is there a place for demo code that will be executed?

I assume you mean is the code in the documentation tested? We use doctests extensively so those can be used for this purpose. I don't insist on them though.

> Should I add tests to verify that the behaviour I am describing in the docs is really what happens?

If you feel like they'd be valuable and feel confident enough to write them. Otherwise I'd skip em.

> I'm not sure how to run the doctests, particularly in the docs/ directory.

I believe our tox setup can run the docs. Otherwise, I go into the docs directory and run `make doctest`.

PR #532, opened 2024-04-15
really drop python<=3.7 support

kloczek:
Filter all python code over `pyupgrade --py38-plus`.


Tinche:
I'll take this, but `typing.List` is still a thing in 3.8 ;)

I can rerun when the fix the tests.

Tinche:
Did this myself in a different PR, thanks!

PR #569, opened 2024-08-08
fix: wrong return type

bzoracler:
Fixes the return type to be consistent with the underlying call:

https://github.com/python-attrs/cattrs/blob/5fd0f5b445928064338e6fcdac081e608b51e541/src/cattrs/gen/typeddicts.py#L239-L246


Tinche (approved):

Tinche:
Thanks!

PR #585, opened 2024-10-01
Stop generating nan values in tests to work with latest attrs (round 2)

natsukium:
this is a follow up commit to 96ed9a1c972814c379f9ea8faa3413aacd4ce6cb


Tinche:
Thanks, looks like a file needs to be formatted with black to pass the lint step.

Tinche:
I got it myself, thanks!

natsukium:
Sorry, I missed the notification. Thank you!

PR #594, opened 2024-10-22
Converter parameter type hints

declaresub:
Change type of unstruct_collection_overrides from Mapping[type, Callable] to Mapping[type, UnstructureHookT] in Converter.__init__.


Tinche:
Hi!

I took at look at this. I don't think `UnstructureHookT` is correct though - that's a TypeVar and this isn't a generic function. It should probably be `UnstructureHook` instead. Does that solve your pyright issue?

Also please add a small changelog entry ;)

Tinche (approved):

Tinche:
Thanks!

PR #607, opened 2024-12-02
Let's try branch coverage

Tinche:


salotz:
Branch coverage is so annoying. FWIW I use this [pytest-checklist](https://github.com/examol-corp/pytest-checklist) which I worked on which helps keep track of "unit" coverage. I.e. that every function has a test. Which IMO is of more primary importance than branch coverage.

PR #617, opened 2025-01-06
615-structuring-a-non-dict-type-as-typeddict-raises-attributeerror

Tinche:


bluetech:
I tested this PR and it works great, thanks!

PR #636, opened 2025-03-18
Fix `unstruct_collection_overrides` callback type

layday:
Same as #594, but for the `copy` method.


Tinche:
Cool, thanks. Guess it's too minor for a changelog entry.

PR #642, opened 2025-03-31
Fix reliance on unspecified behavior in test_disambiguators.py

JelleZijlstra:
See python/cpython#131933. Python does not guarantee that running `Union[A, B]` twice gives the same object back. This test is currently broken in Python 3.14 main.

### DIFF
--- a/tests/test_disambiguators.py
+++ b/tests/test_disambiguators.py
@@ -208,7 +208,7 @@ def test_disambiguate_from_discriminated_enum():
 
     fn = create_default_dis_func(c, E, F, G)
     assert fn({"op": 1}) is E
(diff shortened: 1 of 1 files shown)


Tinche:
Cool, thanks. I wonder if there's user code in the wild that relies on this behavior though - cattrs supports matching types by predicate functions, so folks might have working predicates that use `is` instead of `==`.

Taking a broader view, it's also just a little unfortunate in that `int is int` and `list is list`, but `list[int] is not list[int]`.

JelleZijlstra:
Thanks for merging the PR!

> Taking a broader view, it's also just a little unfortunate in that int is int and list is list, but list[int] is not list[int].

Possibly, but I don't think this is terribly inconsistent with how Python works in general. It's also true that `(int, int) is not (int, int)`, and `int | str is not int | str`.

PR #653, opened 2025-06-02
Enable Python 3.14

Tinche:


Tinche:
> Python 3.14.0rc2 adds a [workaround](https://github.com/python/cpython/pull/137666/files) to remove the “bad” `__subclasses__` entries in dataclasses with  `__slots__`.

Sweet, thanks for letting me know. I might still keep the current filter in, since that may allow us to skip the gc in the future. Letting the original classes be garbage collected again is definitely important for third-party user code out in the wild, though.

PR #655, opened 2025-06-04
Fix structuring of attrs class that inherit from typing/collections.abc generic type aliases

### LINKED ISSUE #654: AttributeError: no __parameters__ when structuring subclass of generic alias types from collections.abc/typing
Hello,

since the latest cattrs v25.1.0, I am experiencing an `AttributeError: type object CLASS has no attribute '__parameters__'` when trying to structure attrs classes that are subclasses of one of `collections.abc` or `typing` generic alias types (e.g. `Mapping[str, int]`)

For example:

```python
from collections.abc import Mapping
from attrs import define, fields
from cattrs import Converter
from cattrs.gen import make_dict_structure_fn

@define
class Point(Mapping[str, int]):
    x: int = 0
    y: int = 0
    def __getitem__(self, key):
        try:
            return getattr(self, key)
        except AttributeError:
            raise KeyError(key) from None
    def __iter__(self):
        return (f.name for f in fields(self.__class__))
    def __len__(self):
        return len(fields(self.__class__))

converter = Converter()
    Point,
    make_dict_structure_fn(Point, converter, _cattrs_forbid_extra_keys=True),
)

converter.structure({"x": 2, "y": 3}, Point)
```

I get the following error:

```
  File "<python-input-3>", line 23, in <module>
    make_dict_structure_fn(Point, converter, _cattrs_forbid_extra_keys=True),
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/clupo/oss/cattrs/src/cattrs/gen/__init__.py", line 728, in make_dict_structure_fn
    mapping = generate_mapping(base, mapping)
  File "/Users/clupo/oss/cattrs/src/cattrs/gen/_generics.py", line 39, in generate_mapping
    parameters = origin.__parameters__
                 ^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'Mapping' has no attribute '__parameters__
```

If instead of using a generic base class, I define my example `Point` class above simply as `class Point(Mapping)` without subscripting `Mapping[str, int]` then the cattrs structuring works.

The error did not occur with the previous cattrs release (v24.1.3) so it looks like a regression.

Note this is exactly the same error as the one identified in this old issue:
https://github.com/python-attrs/cattrs/issues/217

That issue was fixed with this PR https://github.com/python-attrs/cattrs/pull/221

However it looks like the current code in `cattrs.gen._generics.generate_mapping` no longer handles the absence of `__parameters__` attribute which is not present in classes from collections.abc (generic alias types).

You do have a test that confirms structuring attrs-classes that in turn subclass from typing/collections.abc classes works, see:

https://github.com/python-attrs/cattrs/blob/0bb472a9517fbb9ebfaa60bc8116c07816b6f072/tests/test_converter_inheritance.py#L44-L88

However in those tests you are only testing with non-generic types.

If I modify the test case by adding e.g. `Iterable[int]` to the parametrized `typing_cls`, then the test fails with the same `AttributeError: type object 'Iterable' has no attribute '__parameters__'`.

I bisected the regression in this specific commit that was meant to "drop dead code", but it turns out this wasn't really "dead":
https://github.com/python-attrs/cattrs/pull/606/commits/2fe721ed291d53fbb538da7c4a846d0ea5676640

anthrotype:
Fixes #654 

The first commit reproduces the error `AttributeError: type object 'Iterable' has no attribute '__parameters__'`.

I will follow up with the actual fix.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,6 +11,11 @@ The third number is for emergencies when we need to start branches for older rel
 
 Our backwards-compatibility policy can be found [here](https://github.com/python-attrs/cattrs/blob/main/.github/SECURITY.md).
 
+## 25.1.1 (UNRELEASED)
+
+- Fixed `AttributeError: no attribute '__parameters__'` while structuring attrs classes that inherit from parametrized generic aliases from `collections.abc`.
+  ([#654](https://github.com/python-attrs/cattrs/issues/654) [#655](https://github.com/python-attrs/cattrs/pull/655))
+
 ## 25.1.0 (2025-05-31)
 
 - **Potentially breaking**: The converters raise {class}`StructureHandlerNotFoundError` more eagerly (on hook creation, instead of on hook use).
--- a/src/cattrs/gen/_generics.py
+++ b/src/cattrs/gen/_generics.py
@@ -36,7 +36,14 @@ def generate_mapping(cl: type, old_mapping: dict[str, type] = {}) -> dict[str, t
     origin = get_origin(cl)
 
     if origin is not None:
-        parameters = origin.__parameters__
+        # To handle the cases where classes in the typing module are using
+        # the GenericAlias structure but aren't a Generic and hence
+        # end up in this function but do not have an `__parameters__`
+        # attribute. These classes are interface types, for example
+        # `typing.Hashable`.
+        parameters = getattr(get_origin(cl), "__parameters__", None)
+        if parameters is None:
+            return dict(old_mapping)
 
         for p, t in zip(parameters, get_args(cl)):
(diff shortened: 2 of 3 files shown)


anthrotype:
yes thanks, I can do in a bit after lunch.

anthrotype:
> Don't forget to add a changelog entry

done

Tinche:
Great job on this. Please add a link to the issue too to the CHANGELOG and I will release this in the evening.

PR #665, opened 2025-07-02
DOCS: Fix version for "default structure hook fallback factory" migration

hoodmane:
The change was moved to version 25.1.0 from version 24.2.0 in #618 but the migration entry was not updated accordingly.


Tinche:
Nice catch!

PR #667, opened 2025-07-03
Do not untructure the Singleton `attrs.NOTHING` at all (identity) instead of to `1` (int) (attrs >= 22.2 problem)

raabf:
Since attrs >= 22.2 the `attrs.NOTHING` is not an object any more but implemented by an `enum.Enum`, [see here](https://github.com/python-attrs/attrs/blob/7b197b1335e336f4cab7257be7de94b2caa24b25/src/attr/_make.py#L61-L78).

That causes the follow up-problem, since cattrs supports unstructuring `enum.Enum`, it will unstructure it to `1` (int).

```python
>>> import cattrs
>>> import attrs
>>> conv = cattrs.Converter()
>>> conv.unstructure(attrs.NOTHING)
1
```

but this is unintended as `attrs.NOTHING`  should not represent an `int` but a special state in an attrs Attribute.

That leads to two problems.

First, `attrs` do set the value of an class attribute to `attrs.NOTHING` in some faulty states. When you now unstructure such a field with cattrs, it suddenly becomes an `1`.

Second, when you create an attrs Attribute (with `attrs.field` or type annotation) with *no default* the resulting object is an [`attrs.Attribute`](https://www.attrs.org/en/stable/api.html#attrs.Attribute) which `attrs.Attribute.default` value is set to `attrs.NOTHING`. So in that context, `attrs.NOTHING` means the attribute does not have any default value.

However, I have now situations where I compare a value of an attribute with its default (i.e. comparing it with  `attrs.Attribute.default`). And if the attribute is an int-field and by chance was set to `1` it is now suddenly equal to its default value although it is defined to have no default at all — which leads to some errors in my case.

Due to `attrs.NOTHING`’s special meaning of being a faulty state or that no default value is set, I think it is better to not unstructure it (identity as I do in this PR) instead of unstructure it to the (unexpected) int-value of `1`.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -28,6 +28,10 @@ Our backwards-compatibility policy can be found [here](https://github.com/python
 - For {class}`cattrs.errors.StructureHandlerNotFoundError` and {class}`cattrs.errors.ForbiddenExtraKeysError` 
   correctly set {attr}`BaseException.args` in `super()` and hence make them pickable. 
   ([#666](https://github.com/python-attrs/cattrs/pull/666))
+- Do not Unstructure the faulty state {data}`attrs.NOTHING` instead of to `1` (int). The unstructure to `1` was caused
+  by a change in attrs 22.2 that {data}`attrs.NOTHING` is internally represented as an integer {class}`enum.Enum`, 
+  but as it is logically just a Singleton we apply now the identity-function on unstructuring it. 
+  ([#667](https://github.com/python-attrs/cattrs/pull/667))
 
 ## 25.1.1 (2025-06-04)
 
--- a/src/cattrs/converters.py
+++ b/src/cattrs/converters.py
@@ -11,6 +11,7 @@ from inspect import signature as inspect_signature
 from pathlib import Path
 from typing import Any, Optional, Tuple, TypeVar, overload
 
+from attr._make import _Nothing
 from attrs import Attribute, resolve_types
 from attrs import has as attrs_has
 from typing_extensions import Self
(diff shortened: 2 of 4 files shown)


Tinche:
I don't really think this belongs in cattrs, especially since it's very easily overridden by users and I've never heard anyone else run into this issue.

Which states cause attrs to assign NOTHING to instance variables? I'm curious since I maintain attrs as well.

> However, I have now situations where I compare a value of an attribute with its default (i.e. comparing it with attrs.Attribute.default). And if the attribute is an int-field and by chance was set to 1 it is now suddenly equal to its default value although it is defined to have no default at all — which leads to some errors in my case.

Yeah, once we have the singleton PEP we can handle this more gracefully. In the meantime, you can probably have a code path that checks if `default is NOTHING` to handle this case.

Tinche:
> To underline what I mean, I expect that a comparison of an attributes-field with its default value only returns True when it is indeed equal with its default, but this is not the case:

Ah, but this is simply not how it's designed or implemented at the moment. The logical type of `Attribute[T].default` is, essentially, `Literal[NOTHING] | T`. So if you want to check whether `mya1_dict["a1"]` is equal to the default, the proper way of doing this would be:

```python
print(
    "a1 unstructured, should be False as there is no default:",
    (
        (mya1_dict["a1"] == conv.unstructure(attribs["a1"].default))
        if attribs["a1"].default is not attrs.NOTHING
        else False
    ),
)
```

and even that logic can be improved a little bit:

```python
print(
    "a1 unstructured, should be False as there is no default:",
    (
        (
            mya1_dict["a1"]
            == conv.unstructure(
                attribs["a1"].default, unstructure_as=attribs["a1"].type
            )
        )
        if attribs["a1"].default is not attrs.NOTHING
        else False
    ),
)
```

In any case, this seems too niche to include in cattrs. I need to weigh contributions against the maintenance burden to maintainers and the complexity burden to users, and this doesn't meet the bar, unfortunately. Thanks in any case!

Issue #669, opened 2025-07-06
`test_errors_pickling` is flaky


AlexWaygood:
Hi! We run the `cattrs` test suite every night in `typing_extensions`'s CI (to make sure that we haven't accidentally broken `cattrs` due to not-yet-released changes on the `typing_extensions` main branch). The `test_errors_pickling` test appears to have become somewhat flaky recently due to an ordering issue:

```pytb
=================================== FAILURES ===================================
___________ test_errors_pickling[ForbiddenExtraKeysError-err_args2] ____________

err_cls = <class 'cattrs.errors.ForbiddenExtraKeysError'>
err_args = ('', <class 'str'>, {'bar', 'foo'})
tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/test_errors_pickling_Forbidden1')

    @pytest.mark.parametrize(
        "err_cls, err_args",
        [
            (StructureHandlerNotFoundError, ("Structure Message", int)),
            (ForbiddenExtraKeysError, ("Forbidden Message", int, {"foo", "bar"})),
            (ForbiddenExtraKeysError, ("", str, {"foo", "bar"})),
            (ForbiddenExtraKeysError, (None, list, {"foo", "bar"})),
            (
                BaseValidationError,
                ("BaseValidation Message", [ValueError("Test BaseValidation")], int),
            ),
            (
                IterableValidationError,
                ("IterableValidation Msg", [ValueError("Test IterableValidation")], int),
            ),
            (
                ClassValidationError,
                ("ClassValidation Message", [ValueError("Test ClassValidation")], int),
            ),
        ],
    )
    def test_errors_pickling(
        err_cls: type[Exception], err_args: tuple[Any, ...], tmp_path: Path
    ) -> None:
        """Test if a round of pickling and unpickling works for errors."""
        before = err_cls(*err_args)
    
        assert before.args == err_args
    
        with (tmp_path / (err_cls.__name__.lower() + ".pypickle")).open("wb") as f:
            pickle.dump(before, f)
    
        with (tmp_path / (err_cls.__name__.lower() + ".pypickle")).open("rb") as f:
            after = pickle.load(f)  # noqa: S301
    
        assert isinstance(after, err_cls)
>       assert str(after) == str(before)
E       AssertionError: assert 'Extra fields...str: foo, bar' == 'Extra fields...str: bar, foo'
E         
E         - Extra fields in constructor for str: bar, foo
E         ?                                         -----
[...]

PR #670, opened 2025-07-06
Make `ForbiddenExtraKeysError.__str__` deterministic

AlexWaygood:
Fixes https://github.com/python-attrs/cattrs/issues/669. Currently the value returned by this method can differ depending on the iteration order of a set, which isn't deterministic. This causing `test_errors_pickling` to be flaky.


Tinche:
Thanks!

PR #677, opened 2025-08-04
Added missing imports to hook factory decorator example.

carltongibson:
Hey @Tinche 👋 

I was just reviewing the docs, and noticed the `get_args` import missing, which I think it adds to have. I put the example in the editor and it complained about `Any` and `Callable` too (which I guess is fair). I thought I'd send in the clean up.


Tinche:
Nice, thanks! Let's be modern and import `Callable` from `collections.abc` though ;)

carltongibson:
> Let's be modern...

You kids 😉

Tinche:
Gracias!

Issue #682, opened 2025-09-02
include_subclasses with configure_tagged_union fails for generic classes


danarmak:
This is the same as #648, except I'm using the tagged union strategy. [test_parents_with_generics](https://github.com/python-attrs/cattrs/blob/727aa894d48bbd31ed1923627ed160a90932d077/tests/strategies/test_include_subclasses.py#L419) fails with this strategy:

```
>       cattrs.strategies.include_subclasses(GenericParent[str], converter, union_strategy=cattrs.strategies.configure_tagged_union)
[snip]
E       AttributeError: __subclasses__. Did you mean: '__subclasscheck__'?
```

The problem is that `_include_subclasses_with_union_strategy` -> `_has_subclasses` still calls `cl.__subclasses__()`, and `typing.get_origin(cl)` needs to be used here too.

Fixing _has_subclasses leads to a further error:

```
  File "test_cattrutil.py", line 111, in test_test
    cattrs.strategies.include_subclasses(GenericParent[str], converter, union_strategy=cattrs.strategies.configure_tagged_union)
  File "cattrs/strategies/_subclasses.py", line 84, in include_subclasses
    _include_subclasses_with_union_strategy(
  File "cattrs/strategies/_subclasses.py", line 234, in _include_subclasses_with_union_strategy
    subclasses = tuple([c for c in union_classes if issubclass(c, cl)])
                                                    ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/typing.py", line 1225, in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
TypeError: Subscripted generics cannot be used with class and instance checks

```

Changing this to read `subclasses = tuple([c for c in union_classes if issubclass(typing.get_origin(c) or c, typing.get_origin(cl) or cl)])` works. This extended test succeeds - note that there is no reason to specify Parent[str] anymore; we don't have to know what type arguments Parent can ever gets, as long as every child class has no open type parameters left:

```
converter = cattrs.Converter()

@define
class GenericParent[T]:
    p: T

@define
class Child1G(GenericParent[str]):
    c: str

@define
class Child2G(GenericParent[int]):
    c: str

cattrs.strategies.include_subclasses(GenericParent[Any], converter, union_strategy=cattrs.strategies.configure_tagged_union)

assert converter.structure(converter.unstructure(Child1G('a', 'b')), GenericParent[Any]) == Child1G('a', 'b')
[...]

PR #650, opened 2025-05-22
Support generic parents in include_subclasses strategy

zhukovgreen:
Resolves #648

The strategy was using parent_cls.__subclasses__() to get the list of subclasses. In case of generics, this method is unavailable.

The fix applies sanitizing the cl and getting its origin class for getting a sublcasses tree.

The class itself remains generic in the tree.

PR #683, opened 2025-09-02
Fix include_subclasses strategy with a generic parent and tagged_union

danarmak:
This extends the fix in #650 to the tagged union strategy. I reported the issue in #682.

Without this, using include_subclasses with tagged_union and a generic parent class raises the error `AttributeError: __subclasses__. Did you mean: '__subclasscheck__'?`.

Note: I don't know whether to edit `default_tag_generator` to say `return (typing.get_origin(typ) or typ).__name__`. It doesn't break any tests, but I'm not sure if there are cases where it wouldn't be appropriate, or what else ought to be changed if this is. For now I left it as a tag_generator override at the callsite, but it would be better for users not to have to pass this. Please advise.

The tests don't pass for me locally on the main branch (regardless of my change), but the failure is in a test that doesn't use these strategies and all the tests passed when I rebased my fix on 25.1.1, so I hope I'm just confused about something unrelated. (The failure is `tests/test_gen_dict.py::test_individual_overrides - assert 'Factory' not in "{'a': ('', ... '_b': None}"`)

(ETA: the tests now pass, see comments)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -17,6 +17,8 @@ Our backwards-compatibility policy can be found [here](https://github.com/python
   ([#684](https://github.com/python-attrs/cattrs/pull/684))
 - Make some Hypothesis tests more robust.
   ([#684](https://github.com/python-attrs/cattrs/pull/684))
+- {func} `cattrs.strategies.include_subclasses` now works with generic parent classes and the tagged union strategy.
+  ([#683](https://github.com/python-attrs/cattrs/pull/683))
 
 ## 25.2.0 (2025-08-31)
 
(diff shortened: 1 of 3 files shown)


Tinche:
Thanks, I'll take a look when I get some free cycles. Can you get me a more detailed error message on that flaky test in the meantime?

Tinche:
Cool, thanks. Can you rebase, I reworked those tests.

Tinche:
Looks great to me. I'm away from a computer until tomorrow, what happens if we don't substitute the tag generator?

PR #696, opened 2025-11-04
Apply attrs' converter to default before omit_if_default check

antony-frolov:


antony-frolov:
Hi @Tinche! I've added a test to fix coverage, but don't really know what are the next steps to get this merged, can I get a review from you?

Tinche:
Yep, I've been busy but this is on my mind to review soon!

src/cattrs/gen/__init__.py
```diff
@@ -177,16 +177,27 @@ def make_dict_unstructure_fn_from_attrs(
...
+            if c is not None:
+                conv_name = f"__c_conv_{attr_name}"
+                globs[conv_name] = c
+                internal_arg_parts[conv_name] = c
+                if isinstance(c, Converter):
+                    field_name = f"__c_field_{attr_name}"
+                    globs[field_name] = a
+                    internal_arg_parts[field_name] = a
+                    def_str = f"{conv_name}({def_str}, instance, {field_name})"
+                else:
```
Tinche: If the default is not a `Factory`, and the converter is not a `Converter`, we can calculate the default at hook generation time (as opposed to hook run time). It's just `converter(default)`.

Maybe something like:

```python
            c = a.converter
            if c is not None:
                conv_name = f"__c_conv_{attr_name}"
                if isinstance(c, Converter):
                    globs[conv_name] = c
                    internal_arg_parts[conv_name] = c
                    field_name = f"__c_field_{attr_name}"
                    globs[field_name] = a
                    internal_arg_parts[field_name] = a
                    def_str = f"{conv_name}({def_str}, instance, {field_name})"
                elif not isinstance(d, Factory):
                    globs[def_name] = c(globs[def_name])
                else:
                    globs[conv_name] = c
                    internal_arg_parts[conv_name] = c
                    def_str = f"{conv_name}({def_str})"
```

Tinche (changes requested):
Also, we need some docs: please add a `versionchanged` line to the function docstring (you can use NEXT as the version), and a changelog entry.

PR #702, opened 2025-12-02
An initial implementation of unstructuring and structuring enums with complex values

### LINKED ISSUE #699: The value of enums should be unstructured and structured too
I'm running into an interesting issue where the value of an enum is a structured value which isn't being properly unstructured and that's causing issues with serialization.

Here's a simple example:
```
import attrs
import cattrs
import enum

class E(enum.Enum):
    A = 0
    B = 1

class EE(enum.Enum):
    A0 = (E.A, 0)
    A1 = (E.A, 1)
    B0 = (E.B, 0)
    B1 = (E.B, 1)

@attrs.define
class C:
    e: E
    ee: EE

c = C(e=E.B, ee=EE.A1)
print(cattrs.unstructure(c))
```
The `e` attribute of the `C` class is properly unstructured to a `1` value and the `ee` attribute is unstructured to the tuple `(E.A, 1)`, but the `E.A` in that resulting tuple is not unstructured, which causes, for instance, JSON serialization to fail.

reversefold:
Fixes #699. The values of an Enum are unstructured and structured. Structuring is only done when the Enum class has a type defined for the `_value_` attribute as per [the Python typing standard](https://typing.python.org/en/latest/spec/enums.html#member-values).

I've added two very simple tests which show that this works. I'm attempting to add tests which use hypothesis like the rest but the magic chains of incantations in `tests/typed.py` are throwing me a bit and it's taking a while to understand what they're doing and how to use it to generate random dynamic Enum classes. Any pointers there would be appreciated.

### DIFF
--- a/src/cattrs/converters.py
+++ b/src/cattrs/converters.py
@@ -9,7 +9,7 @@ from enum import Enum
 from inspect import Signature
 from inspect import signature as inspect_signature
 from pathlib import Path
-from typing import Any, Optional, Tuple, TypeVar, get_type_hints, overload
+from typing import Any, Optional, Tuple, TypeVar, overload
 
 from attrs import Attribute, resolve_types
 from attrs import has as attrs_has
--- a/tests/test_enums.py
+++ b/tests/test_enums.py
@@ -38,6 +38,7 @@ class SimpleEnum(Enum):
     B = 1
     C = 2
 
+
 class SimpleEnumWithTypeHint(Enum):
     _value_: str
     D = "D"


src/cattrs/converters.py
```diff
@@ -713,6 +713,16 @@ def _structure_simple_literal(val, type):
             raise Exception(f"{val} not in literal {type}")
         return val
 
+    def _structure_enum(self, val: Any, cl: type[Enum]) -> Enum:
+        """Structure ``val`` if possible and return the enum it corresponds to.
+
+        Uses type hints for the "_value_" attribute if they exist to structure
+        the enum values before returning the result."""
+        hints = get_type_hints(cl)
```
Tinche: `get_type_hints` is probably going to be expensive. Let's try doing a simple check first. Maybe something like:

```
if "_value_" in cl.__dict__:
    val = self.structure(val, get_type_hints(cl)["_value_"])
```
I want to avoid doing too much extra work in the majority of cases.

src/cattrs/converters.py
```diff
@@ -631,8 +631,8 @@ def unstructure_attrs_astuple(self, obj: Any) -> tuple[Any, ...]:
         return tuple(res)
 
     def _unstructure_enum(self, obj: Enum) -> Any:
-        """Convert an enum to its value."""
-        return obj.value
+        """Convert an enum to its unstructured value."""
+        return self._unstructure_func.dispatch(obj.value.__class__)(obj.value)
```
Tinche: Hm, I'm a little concerned about performance and backwards comp. Can we do a deep unstructure only if the enum has a `_value_` attribute, like when structuring?

docs/defaulthooks.md
```diff
@@ -70,6 +69,30 @@ This works even for complex values, like tuples.
 'siamese'
 ```
 
+Enum structuring and unstructuring even works for complex values, like tuples, but if you have anything but simple literal types in those tuples (`str`, `bool`, `int`, `float`) you should consider defining the Enum value's type via the `_value_` attribute's type hint so that cattrs can properly structure it.
```
Tinche: Let's link https://typing.python.org/en/latest/spec/enums.html#member-values in here somewhere.

Tinche (changes requested):
Sorry for the delay, I'm a little under the weather.

Looks like good work, left some comments.

src/cattrs/converters.py
reversefold: Same as above. Change made so it's conditional like the above as well.

reversefold:
> Sorry for the delay, I'm a little under the weather.
> 
> Looks like good work, left some comments.

No worries on the delay. Comments should be addressed in the additional commits.

Tinche:
LGTM, thanks.

I might work on this a little more before releasing the next version, to see if I can convert this code into a hook factory for more speed. I also might not, time-depending.

PR #723, opened 2026-02-20
Fix wrong parameter name in namedtuple_dict_structure_factory

bysiber:
`namedtuple_dict_structure_factory` passes `_cattrs_use_detailed_validation` to `make_dict_structure_fn_from_attrs`, but the expected parameter name is `_cattrs_detailed_validation` (without the `use_` prefix).

Since `make_dict_structure_fn_from_attrs` accepts `**kwargs` for attribute overrides, the misnamed keyword is silently swallowed as if it were an attribute override key, and `_cattrs_detailed_validation` falls back to its default `"from_converter"`.

This means the `detailed_validation` argument to `namedtuple_dict_structure_factory` has no effect — the converter's default is always used regardless of what's passed.

The fix renames the keyword to match the actual parameter name.

### DIFF
--- a/tests/test_tuples.py
+++ b/tests/test_tuples.py
@@ -11,7 +11,7 @@ from cattrs.cols import (
     namedtuple_dict_unstructure_factory,
 )
 from cattrs.converters import Converter
-from cattrs.errors import ForbiddenExtraKeysError
+from cattrs.errors import ClassValidationError, ForbiddenExtraKeysError
 
 
 def test_simple_hetero_tuples(genconverter: Converter):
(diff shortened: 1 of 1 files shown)


bysiber:
Added a regression test and a HISTORY.md entry in the latest push. The test creates a converter with `detailed_validation=False` but passes `detailed_validation=True` to the factory, verifying the parameter is actually forwarded correctly.

Tinche:
Thanks, I took the liberty of fixing the test and merging the changelog to nudge this along.

PR #724, opened 2026-02-20
Return factory from hook factory decorator functions

bysiber:
`register_unstructure_hook_factory` and `register_structure_hook_factory` can be used as decorators (documented since 24.1.0), but the inner `decorator` function doesn't return the factory. This means:

```python
@converter.register_structure_hook_factory(has)
def my_factory(type):
    ...

my_factory(SomeClass)  # TypeError: 'NoneType' is not callable
```

The type annotations already declare the correct contract (`Callable[[Factory], Factory]`), and internal uses discard the return value, so this only affects external users who use the decorator syntax and later reference the decorated function.

Fix: add `return factory` at the end of both `decorator` functions.


Tinche:
Ooph good catch. Please add a changelog entry though.

Tinche:
I'm starting to prepare the next release so I got this myself. Thanks!

Issue #742, opened 2026-04-01
Union passthrough does not work when target union uses a new-style type alias as element for passthrough


jhominal:
Hello,

I have just found an issue with the `configure_union_passthrough` strategy, when using the 3.12 `type` alias as an element of a union to structure.

Here is my attempt at an MRE, tested on Python 3.12:

```python
import dataclasses
import typing

import cattrs
import cattrs.strategies


converter = cattrs.Converter()
cattrs.strategies.configure_union_passthrough(str | int, converter)


@dataclasses.dataclass
class DataClass:
    field: str


type NewScalar = str
OldScalar: typing.TypeAlias = str

if __name__ == "__main__":
    data = {"field": "value"}

    v1 = converter.structure(data, OldScalar | DataClass)  # This one works
    print(v1)
    v2 = converter.structure(data, NewScalar | DataClass)  # This one does not work
    print(v2)
```

The line before last raises `cattrs.errors.StructureHandlerNotFoundError: Unsupported type: NewScalar | __main__.DataClass. Register a structure hook for it.`

Tinche:
Got it. We should probably be smart here and support both type aliases and newtypes out of the box. I'll see what I can do in the next few days.

jhominal:
Given the way that new-style type aliases can be nested, I wonder if that would basically require a function that "resolves" union by going through the layers of `TypeAliasType` values.

The thing that worries me here is the way that a bugfix interacts with the way that `cattrs` currently documents [hooks that resolve specifically on a new-style type alias](https://catt.rs/en/stable/defaulthooks.html#type-aliases).

PR #745, opened 2026-04-12
Add support for Collection[X] | Y (for some collection types)

### LINKED ISSUE #743: Implement a strategy to resolve natively Sequence[X] | Y
Hello,

I wonder whether you would be interested in providing in `cattrs` a strategy to allow the structuring of `Sequence[X] | Y`, under the conditions that:

 * both `X` and `Y` can be structured;
 * `Y` is not a `list`, `tuple`, `MutableSequence` or `Sequence`;

I have already written the implementation for my work (my main worry is that my colleagues may not be able to maintain that hook factory long-term if I am gone):

```python
def configure_union_with_single_sequence_dispatch(converter: cattrs.BaseConverter):
    def is_union_containing_single_sequence(target_type: Any) -> bool:
        # If target_type is a new-style alias type, we need to resolve on the __value__ field
        if isinstance(target_type, TypeAliasType):
            target_type = target_type.__value__

        if get_origin(target_type) not in (UnionType, Union):
            return False

        type_args = set(target_type.__args__)
        if len(type_args) < 2 or (len(type_args) == 2 and type(None) in type_args):
            return False

        # Design choice: support the case where tuple[X, Y], list[X], MutableSequence[X] or Sequence[X] appears in the Union
        # We could have an issue in cases where one or more elements are behind new-style type aliases
        sequence_types = (tuple, list, MutableSequence, Sequence)
        sequence_type_args = [t for t in type_args if t in sequence_types or get_origin(t) in sequence_types]
        return len(sequence_type_args) == 1

    def make_structure_union_containing_single_sequence(target_type: Any, /) -> Callable[[Any, Any], Any]:
        # If target_type is a new-style alias type, we need to resolve on the __value__ field
        if isinstance(target_type, TypeAliasType):
            target_type = target_type.__value__

        type_args = set(target_type.__args__)
        sequence_types = (tuple, list, MutableSequence, Sequence)
        sequence_type_arg = next(t for t in type_args if t in sequence_types or get_origin(t) in sequence_types)

        other_type_args = [t for t in type_args if t != sequence_type_arg]
        spillover_type: Any = Union[tuple(other_type_args)] if len(other_type_args) > 1 else other_type_args[0]

        def structure_union_containing_single_sequence(val: Any, _: Any) -> Any:
            if isinstance(val, (tuple, list)):
                return converter.structure(val, sequence_type_arg)
            return converter.structure(val, spillover_type)

        return structure_union_containing_single_sequence

    converter.register_structure_hook_factory(is_union_containing_single_sequence, make_structure_union_containing_single_sequence)
```

jhominal:
fixes #743 

As discussed in the issue, here is the PR.

Compared to my last comment on the issue:

 * I have decided to also include support for non-sequence types `deque`, `set` and `frozenset`. In case you would prefer not supporting these types, I can remove them (I am not sure whether the added support is "worth it" or not.)
 * I have removed any attempt to handle #742, and I expect to apply the fix for #742 to this PR when it has been designed.
 * I have refactored some aspects (parameter names, avoiding parent function scope capture) so that the functions look more similar to the union pass-through strategy.

I must admit, I was not too inspired for tests, there are probably more tests that could be added. The PR is also missing documentation.

### DIFF
--- a/src/cattrs/strategies/_unions.py
+++ b/src/cattrs/strategies/_unions.py
@@ -1,5 +1,6 @@
-from collections import defaultdict
-from typing import Any, Callable, Union
+import collections.abc
+from collections import defaultdict, deque
+from typing import Any, Callable, Union, get_origin
 
 from attrs import NOTHING, NothingType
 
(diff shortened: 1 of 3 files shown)


Tinche:
Just letting you know this is on my radar, I'm feeling under the weather this week so it might have to wait a little

jhominal:
Thank you for commenting about your status!

Do not worry about it - there is really nothing on the line here except a possible improvement to users of `cattrs`.

src/cattrs/strategies/_unions.py
```diff
@@ -282,3 +283,74 @@ def contains_native_union(exact_type: Any) -> bool:
...
+
+        def structure_union_single_collection(
+            val: Any,
+            _: Any,
+            collection_type=collection_type_arg,
+            spillover=spillover_type,
+        ) -> Any:
+            # Design choice: only detect known concrete types as valid source types
+            # That avoids having to blacklist e.g. str or bytes
+            if isinstance(val, (deque, frozenset, list, set, tuple)):
```
Tinche, line 350: I was going to ask if we can test against the exact type we're expecting, but that might be too restrictive. It's very common to structure lists into tuples/frozensets etc. So I think this is perfectly fine.

src/cattrs/strategies/_unions.py
```diff
@@ -282,3 +283,74 @@ def contains_native_union(exact_type: Any) -> bool:
...
+                return converter.structure(val, collection_type)
```
Tinche, line 351: A trick you can do is finding these hooks in the hook factory itself (so in `make_structure_union_single_collection`, instead of in `structure_union_single_collection`) using `converter.get_structure_hook`, and then just using them here.

Tinche (commented):
Looks very solid to me. Please proceed with the changelog entry and some documentation.

PR #748, opened 2026-05-02
Make it possible to skip cbor2 tests

mgorny:
Now that `cbor2` has been rewritten in Rust and the fallback Python implementation was removed, it is no longer possible to use it on all architectures where cattrs are used.  Move the `cbor2` imports into specific tests, so that the test suite can be run again, while skipping the handful of tests that actually need it.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -24,6 +24,8 @@ Our backwards-compatibility policy can be found [here](https://github.com/python
   ([#723](https://github.com/python-attrs/cattrs/pull/723))
 - _cattrs_ is now autoformatted using Ruff.
   ([#732](https://github.com/python-attrs/cattrs/pull/732))
+- Support running the test suite without `cbor2` installed.
+  ([#748](https://github.com/python-attrs/cattrs/pull/748))
 
 ## 26.1.0 (2026-02-18)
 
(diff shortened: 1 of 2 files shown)


mgorny:
I've aimed at a minimal change here. If you prefer something more elaborate, please let me know.

Tinche:
Which platforms are giving you trouble? 

If we do this, we should follow how we handle missing msgspec and orjson, with appropriate skipifs.

mgorny:
> Which platforms are giving you trouble?

Alpha, ARM < v6, HPPA, some of RISC-v and SPRC configurations, i486.

> If we do this, we should follow how we handle missing msgspec and orjson, with appropriate skipifs.

Sure, but what's "appropriate" here? msgspec and orjson are skipped on PyPy right now.

Tinche:
Maybe we can try importing cbor2 and set the flag if it fails?

mgorny:
Sure. Do you want me to adjust the other tests to match? I can also cover the remaining optional dependencies, if you want.

Tinche:
Let's leave the other tests. Please add a very short changelog entry (although I don't think this is super relevant for end users).

Why the `__import__` call instead of just a good old import statement?

mgorny:
> Why the __import__ call instead of just a good old import statement?

So that import checks don't complain that we're importing `cbor2` but not using it (directly).

Tinche (approved):

Tinche:
Thanks!

mgorny:
Thanks!

PR #760, opened 2026-06-24
fix: unstructure enum values recursively through the converter

### LINKED ISSUE #679: An enum used as another enum's value is not unstructured by the json converter
Trivial example:
```
import cattrs
import enum
import cattr.preconf.json

c = cattr.preconf.json.make_converter()
class TE(enum.Enum):
    A = 1
class TE2(enum.Enum):
    C = TE.A
print(repr(c.unstructure(TE2.C)))
```
Outputs:
```
<TE.A: 1>
```
Should output:
```
1
```
The actual use-case I have for this is an enum with a value that has a tuple with another enum's value and an additional integer value so it isn't quite this trivial. Nevertheless, the problem is the same. An enum's value may be one which needs to be recursively unstructured. In fact, the same issue arises if an `attrs` class instance is used as an enum's value:
```
import attrs
@attrs.define
class C:
    a: int
class EC(enum.Enum):
    D = C(1)
print(repr(c.unstructure(EC.D)))
```
The reverse issue also happens when trying to use `structure` on what the jsonable result should be:
```
c.structure({"a": 1}, EC)
```
```
c.structure(1, TE2)
```
Both fail.

binggao1230:
## Summary

Fixes #679.

When an enum's members have other enums as values (no `_value_` annotation),
`enum_unstructure_factory` returned `e.value` directly, bypassing the converter.
This meant the inner enum was never unstructured—it was returned as a raw `Enum`
object, which causes `json.dumps` (and other serializers) to fail.

**Before:**
```python
class Color(Enum):
    RED = 1

class Style(Enum):
    NORMAL = Color.RED

c = make_converter()          # json converter
c.unstructure(Style.NORMAL)   # → Color.RED  ← raw Enum, not JSON-serializable
```

**After:**
```python
c.unstructure(Style.NORMAL)   # → 1  ✓
```

## Root cause

`enum_unstructure_factory` had two branches:
- typed enum (`_value_` annotation): called `converter.unstructure(e.value)` — correct
- untyped enum: returned `lambda e: e.value` — skipped the converter entirely

The distinction is unnecessary for unstructuring. The fix collapses both branches to
always call `converter.unstructure(e.value)`.

Primitive enum values (int, str) pass through the converter unchanged, so existing
behavior is preserved.

## Test plan

- [x] Existing enum tests (`tests/test_enums.py`) all pass
- [x] New regression test `test_unstructure_enum_with_enum_values` added and passing
- [x] All unstructure and disambiguator tests pass (40 total)

### DIFF
--- a/src/cattrs/enums.py
+++ b/src/cattrs/enums.py
@@ -1,9 +1,24 @@
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from enum import Enum
 from typing import TYPE_CHECKING, Any
 
+from ._compat import has
+
 if TYPE_CHECKING:
     from .converters import BaseConverter
+
+
+def _needs_recursive_unstructure(value: Any) -> bool:
+    if isinstance(value, Enum) or has(value.__class__):
+        return True
+    if isinstance(value, tuple | list | set | frozenset):
+        return any(_needs_recursive_unstructure(v) for v in value)
+    if isinstance(value, Mapping):
+        return any(
+            _needs_recursive_unstructure(k) or _needs_recursive_unstructure(v)
+            for k, v in value.items()
+        )
+    return False
 
 
 def enum_unstructure_factory(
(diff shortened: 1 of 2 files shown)


Tinche:
That's what I was afraid of - by always calling unstructure, we tank the very common hot path for simpler enums. I think the factory should be smarter, and default to unstructure only under certain cases.

binggao1230:
Thanks, agreed. I updated the factory so simple enum values keep the direct `e.value` hook, and only typed enums or enum values that visibly need recursive unstructuring use `converter.unstructure(e.value)`.

The recursive path now covers the cases from #679: enum-valued members, tuples/collections/mappings containing enums, and attrs/dataclass values. I also added a test that registers an overriding `int` unstructure hook and verifies a simple enum still returns the raw value directly.

Verification run locally:

```text
uv run --group test python -m pytest tests/test_enums.py -q
uv run --group lint ruff check src/cattrs/enums.py tests/test_enums.py
uv run --group lint ruff format --check src/cattrs/enums.py tests/test_enums.py
git diff --check
uv run --group test --all-extras python -m pytest tests/test_preconf.py -q
uv run --group test --all-extras python -m pytest bench/test_enums.py --benchmark-disable -q
```

PR #764, opened 2026-06-26
Add Decimal support to json and pyyaml preconf converters

### LINKED ISSUE #761: decimal.Decimal not handled in preconf converters (json, pyyaml)
### Problem

`decimal.Decimal` is a very common stdlib type in financial and e-commerce applications, but none of the preconf converters handle it. Both `unstructure` and `structure` operations fail.

### Reproduction

```python
import decimal
import attrs
from cattrs.preconf.json import make_converter

@attrs.define
class Order:
    amount: decimal.Decimal

conv = make_converter()
o = Order(amount=decimal.Decimal('19.99'))

# Unstructure passes through the raw Decimal object — not JSON serializable
conv.dumps(o)
# → TypeError: Object of type Decimal is not JSON serializable

# Structure fails entirely
conv.structure({'amount': '19.99'}, Order)
# → StructureHandlerNotFoundError: Unsupported type: <class 'decimal.Decimal'>. Register a structure hook for it.
```

Same crash with the `pyyaml` preconf:

```python
from cattrs.preconf.pyyaml import make_converter
conv = make_converter()
conv.dumps(Order(amount=decimal.Decimal('19.99')))
# → RepresenterError: cannot represent an object: Decimal('19.99')
```

### Expected Behavior

The json and pyyaml preconf converters should handle `decimal.Decimal` out of the box:

- **Unstructure**: `Decimal → str` (preserving full precision — NOT `float`, which loses precision for values like `Decimal('0.1')`)
- **Structure**: `str → Decimal`

```python
conv.dumps(Order(amount=Decimal('19.99')))
# → '{"amount": "19.99"}'

conv.loads('{"amount": "19.99"}', Order)
# → Order(amount=Decimal('19.99'))
```

### Why `str` and not `float`

Using `float(Decimal('19.99'))` introduces floating-point precision loss:

```python
>>> float(Decimal('19.99'))
19.99  # looks ok
>>> float(Decimal('0.1')) + float(Decimal('0.2'))
0.30000000000000004  # precision lost
```

Serializing as a string preserves the exact value and is the standard practice for monetary amounts.

### Suggested Fix

In `cattrs/preconf/json.py` and `cattrs/preconf/pyyaml.py` `configure_converter`:

```python
from decimal import Decimal

converter.register_unstructure_hook(Decimal, str)
converter.register_structure_hook(Decimal, lambda v, _: Decimal(v))
```

### Environment

- `cattrs` latest
- Python 3.12

Binilkks:
Closes #761

## Summary

- Registers `Decimal` unstructure/structure hooks in `preconf/json.py` and `preconf/pyyaml.py`
- Decimals are serialized as strings to avoid the float precision loss that `json.dumps(float(Decimal("19.99")))` causes
- pyyaml hook uses `Decimal(str(v))` so that values YAML auto-parses as floats are still re-converted correctly
- Adds 3 new tests in `tests/test_preconf.py`: round-trip, precision preservation, and `dumps`/`loads` integration

## Behaviour

```python
from decimal import Decimal
from attrs import define
from cattrs.preconf.json import make_converter

@define
class Order:
    amount: Decimal

converter = make_converter()
order = Order(amount=Decimal("19.99"))

data = converter.unstructure(order)   # {"amount": "19.99"}  ← string, not float
assert data["amount"] == "19.99"

restored = converter.structure(data, Order)
assert restored == order              # exact Decimal equality
```

## Test plan

- [ ] `test_decimal_round_trip` — verifies string representation and round-trip equality for both json and pyyaml converters
- [ ] `test_decimal_precision_preserved_json` — `0.1 + 0.2` round-trips without float precision loss
- [ ] `test_decimal_json_dumps_loads` — end-to-end `dumps`/`loads` with multiple Decimal fields

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cattrs/preconf/json.py
+++ b/src/cattrs/preconf/json.py
@@ -3,6 +3,7 @@
 from base64 import b85decode, b85encode
 from collections.abc import Set
 from datetime import date, datetime
+from decimal import Decimal
 from json import dumps, loads
 from typing import Any, TypeVar, Union
 
@@ -32,6 +33,7 @@ def configure_converter(converter: BaseConverter) -> None:
 
     * bytes are serialized as base85 strings
     * datetimes are serialized as ISO 8601
+    * decimals are serialized as strings to preserve precision
     * counters are serialized as dicts
     * sets are serialized as lists
     * string and int enums are passed through when unstructuring
@@ -45,6 +47,8 @@ def configure_converter(converter: BaseConverter) -> None:
         bytes, lambda v: (b85encode(v) if v else b"").decode("utf8")
     )
     converter.register_structure_hook(bytes, lambda v, _: b85decode(v))
+    converter.register_unstructure_hook(Decimal, str)
+    converter.register_structure_hook(Decimal, lambda v, _: Decimal(v))
     converter.register_unstructure_hook(datetime, lambda v: v.isoformat())
     converter.register_structure_hook(datetime, lambda v, _: datetime.fromisoformat(v))
     converter.register_unstructure_hook(date, lambda v: v.isoformat())
--- a/src/cattrs/preconf/pyyaml.py
+++ b/src/cattrs/preconf/pyyaml.py
@@ -1,6 +1,7 @@
 """Preconfigured converters for pyyaml."""
 
 from datetime import date, datetime
+from decimal import Decimal
 from functools import partial
 from typing import Any, TypeVar, Union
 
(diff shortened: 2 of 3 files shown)


Tinche:
I think this is a good change. Can we also apply it to the rest of the json preconfs? Ujson, orjson and msgspec. Note that some of them may already be doing the right thing, in which case we should let them handle it by default.

PR #765, opened 2026-07-06
Disambiguate unions independent of member class order (#230)

### LINKED ISSUE #230: disambiguators.create_uniq_field_dis_func fails to disambiguate unions depending on parameter order
* cattrs version: 1.10.0
* Python version: 3.10.0
* Operating System: Win

### Description

In some cases, `create_uniq_field_dis_func` fails to disambiguate certain unions. The order of arguments matter, which can cause some confusion if the arguments are taken from a set with no strict ordering, for example, because it sometimes works, and sometimes fails.

Consider the following program:
```py
import attrs, cattrs

# the fields are named after the classes they share
@attrs.define
class A:
    ab: int
    a: str

@attrs.define
class B:
    ab: int
    bc: int

@attrs.define
class C:
    bc: int

cattrs.disambiguators.create_uniq_field_dis_func(A, B, C)  # no error
cattrs.disambiguators.create_uniq_field_dis_func(B, A, C)  # ValueError: <class '__main__.B'> has no usable unique attributes.
```

Another case where this can bite people is if `Union[A, B, C]` was defined, and the code works, then later on, somebody defines `Union[B, A, C]`. Python will convert the types into whatever was defined first, so suddenly, `Union[A, B, C]` is treated like `Union[B, A, C]`

vineethsaivs:
Fixes #230.

### Problem

`create_default_dis_func` (aka `create_uniq_field_dis_func`) fails to disambiguate a valid union depending on the order its member classes are passed in.

```python
from attrs import define
from cattrs import Converter
from cattrs.disambiguators import create_default_dis_func

@define
class A: ab: int; a: str
@define
class B: ab: int; bc: int
@define
class C: bc: int

c = Converter()
create_default_dis_func(c, A, B, C)   # ok
create_default_dis_func(c, B, A, C)   # TypeError: <class 'C'> has no usable non-default attributes
```

A valid assignment exists (`A` -> `a`, `B` -> `ab`, `C` -> `bc`), and 3 of the 6 orderings find it while the other 3 raise.

### Root cause

The "unique keys" branch resolves discriminator fields in a single fixed-order pass. For each class it subtracts the fields of every *not-yet-pinned* class and takes what's left as that class's unique fields. A class whose fields are all shared with another still-unpinned class therefore gets no unique field and is pushed into the single `fallback` slot; a second such class then raises. Since the processing order is fixed up front (sorted once by field count) and never re-selects which class to pin next, whether the valid assignment is found depends on the input order. In the example, if `C` is processed while `B` is still unpinned, `C`'s only field `bc` is shared with `B`, so `C` has no unique field even though pinning `B` on `ab` first would free `bc` for `C`.

### Fix

Resolve unique fields iteratively to a fixpoint. On each pass, pin every class that has a non-default field unique among the classes *not yet pinned*. Pinning a class removes its fields from consideration, which can make a previously ambiguous class uniquely identifiable, so passes repeat until one pins nothing. The single class that still cannot be pinned becomes the fallback; more than one remaining is a genuine ambiguity and still raises `TypeError`. This is order-independent and contained to the one function; the discriminator-literal path and both `dis_func` variants are unchanged.

The generated `dis_func` still checks keys in insertion (pin) order, and that stays correct: a class is pinned only on a field unique among the classes still unpinned on that pass, so its key cannot appear in the payload of any class pinned later or in the fallback, and an earlier-pinned class's key cannot appear in a later class's payload.

### Test

`test_disambiguation_is_order_independent` builds the union above and asserts that all 6 orderings both succeed and route each member's payload back to the correct class. It fails on the current code (3 of 6 orderings raise) and passes with this change. The existing `test_edge_errors` cases (no fields, shared single field, unhelpful discriminator, and the fallback cases) still pass, as do the union suites.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,8 @@ Our backwards-compatibility policy can be found [here](https://github.com/python
 
 ## NEXT (UNRELEASED)
 
+- Fix `create_default_dis_func <cattrs.disambiguators.create_default_dis_func>` (aka `create_uniq_field_dis_func`) failing to disambiguate valid unions depending on the order of the member classes; unique fields are now resolved iteratively to a fixpoint.
+  ([#230](https://github.com/python-attrs/cattrs/issues/230) [#765](https://github.com/python-attrs/cattrs/pull/765))
 - Support more recursive types on 3.14+ with specialized factories for [`annotationlib.ForwardRef`](https://docs.python.org/3/library/annotationlib.html#annotationlib.ForwardRef).
   ([#740](https://github.com/python-attrs/cattrs/issues/740) [#741](https://github.com/python-attrs/cattrs/pull/741))
 - `Converter <cattrs.Converter>` now uses specialized hook factories to generate hooks for tuples, increasing speed.
--- a/src/cattrs/disambiguators.py
+++ b/src/cattrs/disambiguators.py
@@ -131,33 +131,47 @@ def create_default_dis_func(
 
     # We start from classes with the largest number of unique fields
     # so we can do easy picks first, making later picks easier.
-    cls_and_attrs.sort(key=lambda c_a: len(c_a[1]), reverse=True)
+    remaining = sorted(cls_and_attrs, key=lambda c_a: len(c_a[1]), reverse=True)
 
-    fallback = None  # If none match, try this.
+    # Iterate to a fixpoint. On each pass, pin every class that has a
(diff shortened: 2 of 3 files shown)


Tinche:
Interesting! LGTM. Will re-review in a couple of days and maybe merge. Thanks.

Tinche:
Alright, let's give this a try. Thanks!

Issue #601, opened 2024-11-20
Enum misuse causes unhelpful error


AstraLuma:
* cattrs version: 24.1.2
* Python version: 3.11
* Operating System: Linux

### Description

Use an enum's value as the default of a class (instead of the enum reference) causes an unhelpful backtrace.

tbh, I'm not entirely sure what the fix here is, but based on how much effort I put into finding this mistake, I think there's room for some improvement in developer experience.

### What I Did

This code:

```
class SiteFlavors(enum.Enum):
    SIMPLE = "simple"


@attrs.define
class Site():
    flavor: SiteFlavors = "simple"

attrs.unstructure(Site())
```

Causes this error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/astraluma/code/teahouse/barista/.venv/lib/python3.12/site-packages/cattrs/converters.py", line 238, in unstructure
    return self._unstructure_func.dispatch(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<cattrs generated unstructure __main__.Site>", line 3, in unstructure_Site
  File "/home/astraluma/code/teahouse/barista/.venv/lib/python3.12/site-packages/cattrs/converters.py", line 359, in _unstructure_enum
    return obj.value
           ^^^^^^^^^
AttributeError: 'str' object has no attribute 'value'
```

which is obvious when reduced, but in my case, the usage was spread out between a definition, usage, and a wrapping library, so it looked more like:

```
  File "/app/src/barista/routes/sites.py", line 44, in delete_site
    await sitedb.attempt_delete(site)
  File "/usr/local/lib/python3.13/site-packages/chaise/__init__.py", line 441, in attempt_delete
    _, db, docid, etag = self._doc2blob(doc)
                         ~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/site-packages/chaise/__init__.py", line 308, in _doc2blob
    blob = self._session.loader().dumpj(doc)
  File "/usr/local/lib/python3.13/site-packages/chaise/__init__.py", line 129, in dumpj
    blob = self.dump_doc(doc)
  File "/usr/local/lib/python3.13/site-packages/chaise/attrs.py", line 194, in dump_doc
    return converter.unstructure(doc)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/site-packages/cattrs/converters.py", line 238, in unstructure
[...]

PR #770, opened 2026-08-14
fix: improve enum misuse error

Sohel2309:
## Bug

Using a non-Enum value with an enum-typed attrs field can result in an
unhelpful `AttributeError: 'str' object has no attribute 'value'`.

## Root cause

`enum_unstructure_factory()` generated unstructure hooks that accessed
`.value` without first checking that the value was an instance of the
expected Enum type.

## Fix

Add an `isinstance()` check for the expected Enum type before accessing
`.value`. Invalid values now raise a clear `TypeError` describing the
expected Enum type and the received value.

The fix covers both the regular Enum and typed-enum paths.

## Tests

Added regression tests covering:
- regular Enum misuse
- typed Enum misuse
- the attrs end-to-end case from issue #601

The regression tests were verified to fail before the fix and pass after
the fix.

## Changelog

Added the fix to `HISTORY.md`.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,8 @@ Our backwards-compatibility policy can be found [here](https://github.com/python
 
 ## NEXT (UNRELEASED)
 
+- Fix unstructuring an enum-typed value that isn't actually an instance of that enum (for example, a raw value assigned directly to an attrs attribute) raising an opaque `AttributeError` instead of a clear, actionable `TypeError`.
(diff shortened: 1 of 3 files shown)


Tinche:
The problem is this inserts an isinstance check into every unstructure call, affecting performance. cattrs expects the type hints to be correct, so I'm not inclined to trade performance for better error messages in this particular case.

PR #771, opened 2026-08-31
Escape rename keys when generating (un)structuring code

winklemad:
## Summary

A field's `rename` target is a dictionary key, so it can legitimately contain any character. But the generated (un)structuring code interpolates it into the source inside single quotes:

```python
res['{kn}'] = ...
instance.x = o['{kn}']
if '{kn}' in o:
```

So a rename to a key containing a quote produces invalid source and blows up when it's compiled:

```python
from attrs import define
from cattrs import Converter
from cattrs.gen import make_dict_unstructure_fn, override

@define
class A:
    b: int

c = Converter()
c.register_unstructure_hook(A, make_dict_unstructure_fn(A, c, b=override(rename="it's")))
c.unstructure(A(1))
# SyntaxError: unterminated string literal (<cattrs generated unstructure __main__.A>)
```

## Fix

Interpolate the key with `repr` (`{kn!r}`) instead of hand-written single quotes, so any key string is embedded as a valid, correctly-escaped literal. This only affects the `rename` key (`kn`); attrs already rejects aliases that aren't valid identifiers, so the alias-based keys can't hit this.

After the change, `it's`, keys with double quotes, and keys that look like code all round-trip as plain string keys.

## Tests

Added `test_renaming_to_key_with_special_characters` (structure + unstructure round-trip for a rename containing a quote and a couple of other awkward keys). It raises `SyntaxError` on `main` and passes here. Full `test_gen_dict.py` / `test_gen.py` stay green (63 passed).

### DIFF
--- a/tests/test_gen_dict.py
+++ b/tests/test_gen_dict.py
@@ -253,6 +253,30 @@ def test_renaming_forbid_extra_keys():
     assert cve.value.exceptions[1].extra_fields == {"c"}
 
 
+def test_renaming_to_key_with_special_characters():
+    """A rename target is a dict key, so it may legitimately contain quotes.
+
+    It used to be interpolated into the generated code inside single quotes, so
+    a key like ``it's`` produced invalid source and raised ``SyntaxError``.
+    """
+
+    @define
+    class A:
+        b: int
+
+    for key in ("it's", 'a"b', "a'] or 1 or ['b"):
+        converter = Converter()
+        converter.register_unstructure_hook(
+            A, make_dict_unstructure_fn(A, converter, b=override(rename=key))
+        )
+        converter.register_structure_hook(
+            A, make_dict_structure_fn(A, converter, b=override(rename=key))
+        )
+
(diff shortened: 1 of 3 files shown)


tests/test_gen_dict.py
```diff
@@ -253,6 +253,30 @@ class A:
...
+
+    It used to be interpolated into the generated code inside single quotes, so
+    a key like ``it's`` produced invalid source and raised ``SyntaxError``.
+    """
+
+    @define
+    class A:
+        b: int
+
+    for key in ("it's", 'a"b', "a'] or 1 or ['b"):
```
Tinche: Let's parametrize the test instead of doing a for loop

Tinche (changes requested):
Looks good, left a comment. Thanks

winklemad:
Done, parametrized in `df3e28f`'s follow-up — thanks!

Tinche (approved):

Tinche:
Great, thanks!

PR #780, opened 2026-09-17
Apply field renames when disambiguating on a Literal field

### LINKED ISSUE #779: Literal discriminator ignores override(rename=...), so union structuring fails with KeyError
When a union's members are discriminated by a `Literal` field, `create_default_dis_func` looks that field up in the payload under its **original** attribute name, ignoring any `override(rename=...)` registered for it. Structuring the union then fails with a bare `KeyError`, even though each member structures fine on its own.

The unique-field path of the same function does honour renames (via `_usable_attribute_names`), so this is an inconsistency between the two disambiguation strategies rather than a general limitation.

## Reproduction

```python
from typing import Literal, Union
from attrs import define
from cattrs import Converter
from cattrs.gen import make_dict_structure_fn, override

@define
class A:
    kind: Literal["a"]
    a_val: int

@define
class B:
    kind: Literal["b"]
    b_val: int

c = Converter()
for cl in (A, B):
    c.register_structure_hook(
        cl, make_dict_structure_fn(cl, c, kind=override(rename="type"))
    )

c.structure({"type": "a", "a_val": 1}, A)            # A(kind='a', a_val=1)  -- fine
c.structure({"type": "a", "a_val": 1}, Union[A, B])  # KeyError: 'kind'
```

## Expected

`A(kind='a', a_val=1)`, the same as structuring `A` directly.

## Actual

```
KeyError: 'kind'
```

## Notes

For contrast, the unique-field strategy handles the equivalent rename correctly — this is what `test_field_renaming` in `tests/test_disambiguators.py` already covers:

```python
@define
class C:
    c_val: int

@define
class D:
    d_val: int

c3 = Converter()
c3.register_structure_hook(C, make_dict_structure_fn(C, c3, c_val=override(rename="cv")))
c3.register_structure_hook(D, make_dict_structure_fn(D, c3, d_val=override(rename="dv")))
c3.structure({"cv": 1}, Union[C, D])  # C(c_val=1) -- correct
```

In `create_default_dis_func`, `overrides` is computed up front but the `use_literals` branch builds its candidate field names from the raw `at.name` and looks up `data[best_discriminator]` with that same raw name; only the later unique-key branch passes `overrides` through `_usable_attribute_names`.

Tested on `main` (5bf7c97). I have a fix and will open a PR shortly.

onk3sh:
Fixes #779.

`create_default_dis_func` computes attribute `overrides` up front, but the `use_literals` branch never applied them: it built its candidate discriminator names from the raw `at.name`, and the generated `dis_func` looked the value up as `data[best_discriminator]` using that same raw name. A union whose `Literal` discriminator field was renamed with `override(rename=...)` therefore failed to structure with a bare `KeyError`, even though each member structured correctly on its own.

The unique-key branch further down already routes names through `_usable_attribute_names`, and `test_field_renaming` covers it. This brings the literal strategy in line.

## Before

```python
@define
class A:
    kind: Literal["a"]
    a_val: int

@define
class B:
    kind: Literal["b"]
    b_val: int

c = Converter()
for cl in (A, B):
    c.register_structure_hook(
        cl, make_dict_structure_fn(cl, c, kind=override(rename="type"))
    )

c.structure({"type": "a", "a_val": 1}, A)            # A(kind='a', a_val=1)
c.structure({"type": "a", "a_val": 1}, Union[A, B])  # KeyError: 'kind'
```

## After

Both return `A(kind='a', a_val=1)`.

## The change

Candidate names are now built from the overridden names, with a per-class `back_map` from the payload name to the original attribute name so the `Literal` args are still read off the right field. The discriminator set is still the intersection across all members, so it only matches when every member agrees on the payload name.

## Testing

`test_renamed_literal_discriminator` added next to the existing `test_field_renaming`. It fails on `main` with `KeyError: 'kind'` across all four converter parametrizations and passes with the fix. Full suite: 994 passed, 15 xfailed, no regressions. `ruff format --check` and `ruff check` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Voyagerroc-Lab (approved):
The Literal discriminator candidates now use the renamed payload key and map that key back to the original attribute to read its Literal values. The regression covers both classes in a renamed union. CI passes across Python 3.10–3.14 and PyPy, including linters and docs. Review performed with OpenAI Codex at the account owners request.

